PR #15, opened 2022-02-12
[wip] Feature/export dialect provider

Kiryous:


examples/jet/pages/fullscreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const toggleTheme = useCallback(() => {
+    setTheme(theme === 'dark' ? 'light' : 'dark');
+  }, [theme]);
+
+  return (
+    <>
+      <DialectProvider
+        wallet={wallet}
+        network={'devnet'}
+        publicKey={DIALECT_PUBLIC_KEY}
```
cbosborn, line 100: wondering if it makes sense to have a `publicKey` in the provider here, since this is specific to a particular dialect use case: when you know the other participant's public key

e.g. this is no longer relevant in my pr here #26, where we're aggregating multiple dialects in a single component. haven't fully resolved what the api should be

examples/jet/pages/fullscreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { WalletContext, Wallet as WalletButton } from '../components/Wallet';
+import BellIcon from './icons/Bell';
+import SettingsIcon from './icons/Gear';
+import BackIcon from './icons/BackArrow';
+
+const DIALECT_PUBLIC_KEY = new anchor.web3.PublicKey(
+  'FkZPdBJMUFQusgsC3Ts1aHRbdJQrjY18MzE7Ft7J4cb4'
+);
+
+export const themeVariables = {
```
cbosborn, line 14: this is a repeat of the styles from the non-full page example, correct?

packages/dialect-react-ui/components/DialectProvider/index.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useEffect } from 'react';
+import * as anchor from '@project-serum/anchor';
+import {
+  connected,
+  useApi,
+  WalletType,
+  ApiProvider,
+  DialectProvider as DialectProviderInternal,
```
cbosborn, line 8: What's the motivation for two providers, one internal?

packages/dialect-react-ui/components/NotificationCenterButton/index.tsx
```diff
@@ -111,25 +76,9 @@ function WrappedNotificationCenterButton(
           // className="w-full h-full bg-white/10"
           // style={{ backdropFilter: 'blur(132px)' }}
         >
-          <NotificationCenter />
+          <NotificationCenter className={popup} />
```
cbosborn, line 79: what's the purpose of this `className`? how does it relate to the themes variables in the provider?

examples/jet/pages/fullscreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+    setTheme(theme === 'dark' ? 'light' : 'dark');
+  }, [theme]);
+
+  return (
+    <>
+      <DialectProvider
+        wallet={wallet}
+        network={'devnet'}
+        publicKey={DIALECT_PUBLIC_KEY}
+        theme={theme}
```
cbosborn, line 101: not specific to this pr but wondering if there's a better way to organize darkmode/lightmode (`theme`) together with the injected styles for those modes (`variables={themeVariables}`). it wasn't immediately clear what each was or why they're different from each other

examples/jet/pages/_app.tsx
```diff
@@ -1,10 +1,28 @@
...
+    <>
+      <Head>
+        <link rel="preconnect" href="https://fonts.googleapis.com" />
+        <link
+          rel="preconnect"
+          href="https://fonts.gstatic.com"
+          crossOrigin="true"
+        />
+        <link
+          href="https://fonts.googleapis.com/css2?family=Poppins:wght@400;500;700&display=swap"
```
cbosborn, line 20: is this here for a jet-specific font?

examples/jet/pages/index.tsx
```diff
@@ -98,29 +98,19 @@ function AuthedHome() {
...
-            publicKey={DIALECT_PUBLIC_KEY}
-            theme={theme}
-            variables={themeVariables}
-          />
-          <WalletButton />
-        </div>
+        <DialectProvider
+          wallet={wallet}
+          network={'devnet'}
+          publicKey={DIALECT_PUBLIC_KEY}
```
cbosborn, line 105: arbitrary naming issue but if this is a jet example it should be called `JET_PUBLIC_KEY`

cbosborn (changes requested):
@Kiryous nice job. made some requests in the review comments. I'm also wondering if we should break this up a bit, into:

1.  pulling the provider out of the `NotificationCenterButton` component, vs
2. implementing a full screen example

asking because I think it makes the most sense to put some more time into a great full screen example, potentially even along as its own example. but that shouldn't block getting the provider separation in, other than vetting that it works

packages/dialect-react-ui/components/DialectProvider/index.tsx
Kiryous: It's to avoid naming collision. We have `ApiProvider` and `DialectProvider` in `@dialectlabs/react` and we want to export a provider from `@dialectlabs/react-ui`, maybe DialectUIProvider is a better naming. What do you think?

examples/jet/pages/fullscreen.tsx
Kiryous: good feedback! One solution could be to rename the `variables` into `themes` and `theme` to `currentTheme` so it will be clear what you need to choose from what in variables dictionary ("light"/"dark")

examples/jet/pages/_app.tsx
Kiryous: Yeah, just for jet example

packages/dialect-react-ui/components/NotificationCenterButton/index.tsx
Kiryous: I just moved the className for notification center popup from inside to better control the styling

Kiryous:
Resolved in 1.0.0-beta

PR #25, opened 2022-02-16
hotfix: full width & height notification center on mobile, remove rel…

Kiryous:
…ative position, calculate top via js

### DIFF
--- a/packages/dialect-react-ui/components/NotificationCenterButton/index.tsx
+++ b/packages/dialect-react-ui/components/NotificationCenterButton/index.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import * as anchor from '@project-serum/anchor';
 import {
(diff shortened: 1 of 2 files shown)


cbosborn:
@Kiryous is there a way we can use css to accomplish this responsiveness goal?

Kiryous:
> @Kiryous is there a way we can use css to accomplish this responsiveness goal?

I dont't think so. If we use css-only, we use a `position: relative` for proper positioning of notification center, but than we cannot set width 100%. So I propose to go with js for positioning of notification center.

@cbosborn Why do you want to avoid this?

@fsher any thought on this?

fsher:
I wonder if this PR is still relevant, @Kiryous @cbosborn? Might be related to the existing [PR](https://github.com/dialectlabs/react/pull/51)

Kiryous:
> I wonder if this PR is still relevant, @Kiryous @cbosborn? Might be related to the existing [PR](https://github.com/dialectlabs/react/pull/51)

the #51 will replace this one, so I close it

PR #26, opened 2022-02-17
feature/chat

cbosborn:
@fsher & @Kiryous this still needs some work, namely:

- [x] support light/dark mode
- [x] break up chat-related code into exportable, consumable library components
- [x] parameterize styles as we have with the notifications module
- [x] add support for thread settings page where you can delete the thread & recover the rent, as you can with notifs
- [ ] ~separate `DialectContext` & e.g. `DialectsContext` for separate use~
- [ ] ~consider consolidating some of the largely-reusable components between notifications & chat. this may be premature though if the UIs end up diverging more~
- [ ] ~enhance routing~
- [ ] ~consolidate `textStyles` with component styles~

but features are largely there, so opening the pr now for tracking & feedback. demo here https://www.loom.com/share/36cea00b5a9b41e99ce01c84d797f7ce


cbosborn:
@fsher @Kiryous this is ready to go. i've crossed out todos above that are out of scope for now, & can be addressed in followup prs (if at all)

PR #35, opened 2022-03-11
feat: NotificationCenter: accept JSON prop for specifying the notification types

Kiryous:
e.g.

```jsx
<NotificationsButton
  wallet={wallet}
  network={'localnet'}
  publicKey={DIALECT_PUBLIC_KEY}
  theme={theme}
  variables={themeVariables}
  notifications={[
    {name: 'Liqudiations', detail: 'Event'},
    {name: 'Collateral health', detail: 'Below 130%'}
  ]}
/>
```

### DIFF
--- a/packages/dialect-react-ui/components/Notifications/index.tsx
+++ b/packages/dialect-react-ui/components/Notifications/index.tsx
@@ -101,11 +101,17 @@ function Settings(props: {
     <>
       <div className="mb-3">
         <h2 className={cs(textStyles.h2, 'mb-1')}>Notifications</h2>
-        {props.notifications?.map((type) => (
-          <ValueRow key={type.name} label={type.name} className={cs('mb-1')}>
-            {type.detail}
-          </ValueRow>
-        ))}
+        {props.notifications
+          ? props.notifications.map((type) => (
+              <ValueRow
+                key={type.name}
+                label={type.name}
+                className={cs('mb-1')}
+              >
+                {type.detail}
+              </ValueRow>
+            ))
+          : 'No notification types supplied'}
       </div>
       <div>
         <h2 className={cs(textStyles.h2, 'mb-1')}>Thread Account</h2>
--- a/packages/dialect-react-ui/components/Notifications/index.tsx
+++ b/packages/dialect-react-ui/components/Notifications/index.tsx
@@ -64,7 +64,6 @@ function CreateThread() {
         Create notifications thread
       </h1>
       <ValueRow
-        highlighted
         label="Rent Deposit (recoverable)"
         className={cs('w-full mb-4')}
       >
@@ -103,13 +102,7 @@ function Settings(props: {
       <div className="mb-3">
         <h2 className={cs(textStyles.h2, 'mb-1')}>Notifications</h2>
         {props.notifications?.map((type) => (
-          <ValueRow
-            key={type.name}
-            label={type.name}
-            highlighted
-            className={cs('mb-1')}
-            colon={false}
-          >
+          <ValueRow key={type.name} label={type.name} className={cs('mb-1')}>
             {type.detail}
           </ValueRow>
         ))}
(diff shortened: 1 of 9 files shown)


Kiryous:
Screenshot 2022-03-11 at 18 56 36

packages/dialect-react-ui/components/Notifications/index.tsx
```diff
@@ -85,38 +90,52 @@ function CreateThread() {
...
   const { colors, textStyles, icons } = useTheme();
 
   return (
     <>
       <div className="mb-3">
-        <p className={cs(textStyles.body, 'mb-1')}>Included event types</p>
-        <ul className={cs(textStyles.bigText, 'list-disc pl-6')}>
-          <li>Welcome message on thread creation</li>
-        </ul>
+        <h2 className={cs(textStyles.h2, 'mb-1')}>Notifications</h2>
```
cbosborn, line 104: We should probably have a "no notifications supplied" or some other kind of text if none are passed, what do you think?

packages/dialect-react-ui/components/Notifications/index.tsx
```diff
@@ -85,38 +90,52 @@ function CreateThread() {
...
+        {props.notifications?.map((type) => (
+          <ValueRow
+            key={type.name}
+            label={type.name}
+            highlighted
+            className={cs('mb-1')}
+            colon={false}
```
cbosborn, line 111: What is the colon for?

cbosborn (approved):
@Kiryous lgtm. just a couple small comments/qs

packages/dialect-react-ui/components/Notifications/index.tsx
Kiryous: I wonder if it make sense for dev of dapp to leave these empty?

packages/dialect-react-ui/components/Notifications/index.tsx
fsher: Sorry didn't quite understand the use-case of the colon. I don't mind leaving it, just don't quite understand it. Could you please provide an example?

packages/dialect-react-ui/components/Notifications/index.tsx
```diff
@@ -85,38 +90,52 @@ function CreateThread() {
...
-        <p className={cs(textStyles.body, 'mb-1')}>Included event types</p>
-        <ul className={cs(textStyles.bigText, 'list-disc pl-6')}>
-          <li>Welcome message on thread creation</li>
-        </ul>
+        <h2 className={cs(textStyles.h2, 'mb-1')}>Notifications</h2>
+        {props.notifications?.map((type) => (
+          <ValueRow
+            key={type.name}
+            label={type.name}
+            highlighted
```
fsher, line 109: I see that almost all ValueRow components now have `highlighted` passed as a prop. Does it make sense to make it a default?
Also, this might (saying "might" because we have to careful) cause a "boolean trap" since it's a styling feature and if we might have variations like "highlighted" "dimmed" "normal" and the conflict between each other (I've only noticed this sorry...)
Not saying that we need to change this now, but please. leave a todo to improve this. Thanks!

packages/dialect-react-ui/components/Notifications/index.tsx
fsher: Also, since it's an optional prop, no need to provide it here explicitly.

fsher (changes requested):
Looks good, added minor comments.

fsher:
Also, I think it would be a good idea to update README's main example as well.

packages/dialect-react-ui/components/Notifications/index.tsx
Kiryous: It seems now it become useless, since in new designs we have only one look of ValueRow

PR #36, opened 2022-03-11
Allow to style notification bubble and popup wrapper

Kiryous:


packages/dialect-react-ui/components/Notifications/Notification.tsx
```diff
@@ -17,15 +17,15 @@ const timeFormatter = new Intl.DateTimeFormat('en-US', {
 });
 
 export const Notification = ({ message, timestamp }: Props) => {
-  const { textStyles } = useTheme();
+  const { textStyles, notificationBubble, notificationTimestamp } = useTheme();
```
cbosborn, line 20: `notificationBubble` is the individual notification message, correct? What do you think about `notificationMessage` as the name?

packages/dialect-react-ui/components/NotificationsButton/index.tsx
```diff
@@ -93,8 +93,7 @@ function WrappedNotificationsButton(
         onClick={() => setOpen(!open)}
       ></IconButton>
       <Transition
-        className="z-50 absolute top-16 w-96 h-96"
-        style={{ width: '29rem', height: '29rem' }}
+        className={popupWrapper}
```
cbosborn, line 96: Ah great 🙏

packages/dialect-react-ui/components/common/ThemeProvider.tsx
```diff
@@ -71,9 +71,13 @@ export type IncomingThemeValues = {
   textArea?: string;
   messageBubble?: string;
   otherMessageBubble?: string;
+  notificationBubble?: string;
+  notificationTimestamp?: string;
+  notificationsDivider?: string;
   bellButton?: string;
   iconButton?: string;
   sendButton?: string;
+  popupWrapper?: string;
```
cbosborn, line 80: Let's use `modal` instead of `popup`. Is that too much work to change now?

cbosborn (changes requested):
@Kiryous lg, just a couple small requests. Let me know if some name changes are too much work & we can isolate in a separate issue/PR

packages/dialect-react-ui/components/common/ThemeProvider.tsx
Kiryous: updated

packages/dialect-react-ui/components/Notifications/Notification.tsx
Kiryous: yeah, good suggestion, renamed

PR #41, opened 2022-03-23
Feature/add esm build

fsher:
Still wip, minor improvements can be made, but ready to be reviewed.


packages/dialect-react-ui/package.json
```diff
@@ -1,13 +1,23 @@
...
-  "types": "./lib/index.d.ts",
+  "main": "./lib/cjs/index.js",
+  "module": "./lib/es/index.js",
+  "types": "./lib/es/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./lib/es/index.js",
+      "require": "./lib/cjs/index.js"
+    },
+    "./styles": "./lib/index.css"
```
Kiryous, line 15: I lean toward ```"./styles.css": "./lib/index.css"``` to be extreamly clear what user imports and to avoid issues with import base on file extention

Kiryous:
@fsher also, is there idea on how to run examples locally now? It seems the previous approach with including them into transpiling not working with esm exports defined.

packages/dialect-react-ui/package.json
fsher: Unfortunately, I think we can't specify the `.css` part, but I can double-check that.

I agree that it's great to be explicit, however I think that styles.css might also be confusing, since it's also a misconception in terms of the actual file.

fsher:
@Kiryous Sorry, I don't quite understand. Nothing really changes here, since if lib folder doesn't exist it will be ignored is it not?

packages/dialect-react-ui/package.json
Kiryous: @fsher Solana use the ".css" for example https://github.com/solana-labs/wallet-adapter/blob/e15287221b90f8d7bee60726d251a5b4c76d243a/packages/ui/react-ui/package.json#L17

fsher: @Kiryous Thank you so much! Perfect, I'll update that now.

fsher: Updated to ./index.css for better file representation

PR #44, opened 2022-03-26
Refactor/styles isolation

Kiryous:


examples/chat/pages/index.tsx
```diff
@@ -8,30 +8,32 @@ import { defaultVariables } from '@dialectlabs/react-ui/components/common/ThemeP
...
+// dt-border-light: #F0F0F0
 // blue: #448EF7
 
 // TODO: Use useTheme instead of explicitly importing defaultVariables
 export const themeVariables: IncomingThemeVariables = {
   dark: {
     bellButton:
-      'w-12 h-12 shadow-xl shadow-neutral-800 border border-neutral-600 hover:shadow-neutral-700',
-    modal: 'rounded-3xl shadow-xl shadow-neutral-900 border border-neutral-800',
+      'dt-w-12 dt-h-12 dt-shadow-xl dt-shadow-neutral-800 dt-border dt-border-neutral-600 hover:dt-shadow-neutral-700',
```
fsher, line 18: Question: does it makes sense to leave the example styles unprefixed, since examples behave as external projects. It would be actually a great test, if we "unprefixed" tailwind styles on top of the prefixed ones (inside react-ui). What do you think?

examples/notifications/styles/globals.css
```diff
@@ -21,11 +21,11 @@ a {
 
 @layer components {
   .wallet-adapter-button.dialect-wallet-button {
-    border: 1px solid white;
-    @apply flex items-center transition ease-in-out duration-300 text-black dark:text-white border border-neutral-300 dark:border-neutral-600 rounded-md py-2 px-6 bg-transparent shadow-md shadow-neutral-300 dark:shadow-neutral-800;
+    dt-border: 1px solid white;
```
fsher, line 24: this property should not be prefixed

packages/dialect-react-ui/index.css
```diff
@@ -1,11 +1,11 @@
 /* Prefixed preflight */
 .dialect {
```
fsher, line 2: do we still need .dialect?

packages/dialect-react-ui/tailwind.config.js
```diff
@@ -7,5 +7,6 @@ module.exports = {
     extend: {},
   },
   plugins: [],
+  prefix: 'dt-',
   important: '.dialect',
```
fsher, line 11: Since dt prefix is here now, we can safely remove the `important` rule here too.

fsher (changes requested):
Well done, sir! Please also update the version and add a new changelog entry, thanks!

examples/chat/pages/index.tsx
Kiryous: Cool idea, to procced with it we need to build styles for lib and examples separately. Could do in the subsequent PR

packages/dialect-react-ui/index.css
Kiryous: preflight is not prefixed, so yeah, unfortunately, we need it

fsher: Sorry, what I meant was that since we have prefixes now, I think we don't need .dialect anymore. Wdyt?

Kiryous: yeah, if I got the question right it's about the next comment. The thing is if we add `.dialect` on the preflight, it increases specificity of the preflight. So for example, if we have in preflight

```css 
.dialect button { 
  padding: 0; 
}
```
and we use class to add padding,

```jsx 
<button className="px-3" />
```

the prefixed intial styles will be applied.

So there are two ways to resolve:
1. Prefix all styles via `important: '.dialect'`
2. Use classes like `dt-button` instead of prefixed preflight 

After a closer look, It seems to me the later is the only way to go. Since the variable overrides classes need the use of `!important` to overcome `.dialect` specifity also.

PR #45, opened 2022-03-28
Feature/email notifications

Kiryous:


packages/dialect-react-ui/components/IconButton/index.tsx
```diff
@@ -23,7 +23,7 @@ export default React.forwardRef(function IconButton(
         className={iconButton}
         onClick={(event) => {
           event.preventDefault();
-          props?.onClick();
+          props?.onClick && props?.onClick();
```
fsher, line 26: Not very pretty but does the trick: `props?.onClick?.();`

packages/dialect-react-ui/components/Notifications/EmailForm.tsx
```diff
@@ -0,0 +1,222 @@
...
+                  onClick={() => {
+                    setEmailEditing(false);
+                  }}
+                >
+                  Cancel
+                </Button>
+                <Button
+                  className={'basis-1/2'}
+                  disabled={email === ''}
+                  onClick={async () => {
```
fsher, line 128: I'd recommend moving this function to a separate variable above, no need to wrap it with useCallback, since this component rerenders every time we type something inside our input.

packages/dialect-react-ui/components/Notifications/EmailForm.tsx
```diff
@@ -0,0 +1,222 @@
...
+                  {isSavingAddress ? 'Saving...' : 'Submit email'}
+                </Button>
+              </div>
+            )}
+
+            {!isChanging && isEmailEditing ? (
+              <Button
+                className={'basis-full'}
+                disabled={email === ''}
+                onClick={async () => {
```
fsher, line 153: Same here.

packages/dialect-react/utils/errors.ts
```diff
@@ -49,6 +51,20 @@ export const noAccount: ParsedErrorData = {
   matchers: ['Account does not exist'],
 };
 
+export const incorrectEmail: ParsedErrorData = {
```
fsher, line 54: Since email has web2 communications, I don't think we need to handle it through matchers, it must have a normalized response, since we control the backend. wdyt?

fsher (changes requested):
Great job, Kirill! Very well done. Left some minor comments, please check.

zhelezkov (changes requested):

packages/dialect-react/utils/errors.ts
Kiryous: good suggestion

packages/dialect-react-ui/components/Notifications/EmailForm.tsx
fsher: done

packages/dialect-react/utils/errors.ts
fsher: added a TODO comment for now, should resolve later on

fsher (approved):

PR #51, opened 2022-04-05
fix: expand modal to fullscreen on mobile for chat and notifications

Kiryous:


packages/dialect-react-ui/components/ChatButton/index.tsx
```diff
@@ -113,7 +112,7 @@ function WrappedChatButton(
           // className="dt-w-full dt-h-full bg-white/10"
           // style={{ backdropFilter: 'blur(132px)' }}
         >
-          <Chat />
+          <Chat toggleModal={() => setOpen((open) => !open)} />
```
fsher, line 115: Question: why did you decide to go with "toggle" rather than usual "onClose" with this?

If I've checked everything correctly, we usually close it everywhere rather than "toggle" it. wdyt?

fsher (changes requested):
Mostly LGTM, have one design question regarding the prop. Would love to hear your thoughts on this

zhelezkov (approved):
looks very solid

packages/dialect-react-ui/components/ChatButton/index.tsx
Kiryous: kinda followed the similar function name on toggleSettings, but in this case I guess your suggestion to change to onClose is better fit. thanks, man!

packages/dialect-react-ui/components/NotificationsButton/index.tsx
```diff
@@ -81,6 +82,17 @@ function WrappedNotificationsButton(
...
   useEffect(() => setRpcUrl(props.rpcUrl || null), [props.rpcUrl, setRpcUrl]);
 
+  const isMobile = useMobile();
+
+  useEffect(() => {
+    // Prevent scrolling of backdrop content on mobile
+    document.documentElement.classList[open && isMobile ? 'add' : 'remove'](
+      'dt-overflow-hidden',
+      'dt-static',
+      'sm:overflow-auto'
```
fsher, line 92: I think you're missing a dt- prefix here

Kiryous: nice catch, thanks!

fsher (approved):
LGTM, sir! Great job!

PR #52, opened 2022-04-06
Fix/Remove flicker on message send

0xavro:
Removes the flicker on message send. We were re-rendering all the messages before. Every time a message would send `messages` would be set as `[]` because `dialect` would be undefined.

flicker-screenshot

### DIFF
--- a/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
@@ -45,8 +45,8 @@ export default function Thread() {
       setYouCanWrite(membersContainCurrentKey);
     }
   }, [
-    dialect?.dialect
-  ])
+    dialect?.dialect,
+  ]);
 
   const disableSendButton =
     text.length <= 0 ||
--- a/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
@@ -1,4 +1,4 @@
-import React, { KeyboardEvent, FormEvent, useState } from 'react';
+import React, { KeyboardEvent, FormEvent, useState, useEffect } from 'react';
 import cs from '../../../../../../utils/classNames';
 import { useApi, useDialect } from '@dialectlabs/react';
 import type { ParsedErrorData } from '@dialectlabs/react';
@@ -15,6 +15,7 @@ export default function Thread() {
 
   const [text, setText] = useState<string>('');
   const [error, setError] = useState<ParsedErrorData | null | undefined>();
+  const [youCanWrite, setYouCanWrite] = useState<boolean>(false);
 
   const handleError = (err: ParsedErrorData) => {
     setError(err);
@@ -35,9 +36,18 @@ export default function Thread() {
         .catch(handleError);
     }
   };
-  const youCanWrite = dialect?.dialect.members.some(
-    (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
-  );
+
+  useEffect(() => {
+    const membersContainCurrentKey = dialect?.dialect.members.some(
+      (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
+    );
+    if (membersContainCurrentKey) {
+      setYouCanWrite(membersContainCurrentKey);
+    }
+  }, [
+    dialect?.dialect
+  ])
+
   const disableSendButton =
     text.length <= 0 ||
     text.length > 280 ||
--- a/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
@@ -39,7 +39,7 @@ export default function Thread() {
 
   useEffect(() => {
     const membersContainCurrentKey = dialect?.dialect.members.some(
-      (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
+      (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1] // is not admin but does have write privilages
     );
     if (membersContainCurrentKey) {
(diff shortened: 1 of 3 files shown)


packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
```diff
@@ -35,9 +36,18 @@ export default function Thread() {
...
   };
-  const youCanWrite = dialect?.dialect.members.some(
-    (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
-  );
+
+  useEffect(() => {
+    const membersContainCurrentKey = dialect?.dialect.members.some(
+      (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
+    );
+    if (membersContainCurrentKey) {
```
fsher, line 44: I believe this `if` statement is unnecessary here, since if the condition remains `false`, react won't rerender the component. So I think that we can simplify this to simply:

```
const membersContainCurrentKey = ...;

setYouCanWrite(membersContainCurrentKey);
```

wdyt?

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
```diff
@@ -35,9 +36,18 @@ export default function Thread() {
         .catch(handleError);
     }
   };
-  const youCanWrite = dialect?.dialect.members.some(
-    (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
-  );
+
+  useEffect(() => {
+    const membersContainCurrentKey = dialect?.dialect.members.some(
+      (m) => m.publicKey.equals(wallet?.publicKey) && m.scopes[1]
```
fsher, line 42: Could you please add a comment about m.scopes[1], a simple explanation why [1] would be sufficient. This will help us understand later on as we develop. Thanks!

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
```diff
@@ -15,6 +15,7 @@ export default function Thread() {
 
   const [text, setText] = useState<string>('');
   const [error, setError] = useState<ParsedErrorData | null | undefined>();
+  const [youCanWrite, setYouCanWrite] = useState<boolean>(false);
```
fsher, line 18: I see that we have been doing this previously, but I think we shouldn't be describing primitive types in generics, since the default value will help us understand what type it is.

What are your thoughts on this? @zhelezkov also prefers explicitly defining the type. If the majority prefers explicitly describing the type, then I'd freeze it as our convention (we should probable define these sooner or later...). Wdyt?

packages/dialect-react/components/DialectContext/index.tsx
```diff
@@ -210,6 +211,18 @@ export const DialectProvider = (props: PropsType): JSX.Element => {
     fetchError?.type,
   ]);
 
+  useEffect(() => {
+    if (messages.length > 0) {
```
fsher, line 215: There are some repeating conditions here, I suggest we simplify this a bit:

Some deps were missing as well, added them to the array too. Please do check that it's working the same way as your current implementation (this was a blind refactor... might've messed it up).

```
useEffect(() => {
  if (!wallet || !dialect?.dialect) {
    setMessages([]);
    return;
  }

  if (messages.length !== dialect.dialect.messages.length || messages.length === 0) {
    setMessages(dialect.dialect.messages);
  }
}, [wallet, dialect?.dialect, messages.length]);
```

Let me know, what you think. Thanks!

fsher (changes requested):
Well done, great catch with the root of the issue.
Left a couple of comments, please let me know what you think. Thanks!

packages/dialect-react/components/DialectContext/index.tsx
0xavro: Actually, I think it can be as simple as this:

```
useEffect(() => {
    const hasNewMessage = wallet && dialect?.dialect && messages.length !== dialect.dialect.messages.length;
    if (hasNewMessage) {
      setMessages(dialect.dialect.messages);
    }
  }, [
    wallet,
    dialect?.dialect,
    messages.length,
  ]);
```
We just want to update the messages if `dialect.dialect.messages.length` differs from our current `messages.length` but we don't want to update it if `dialect` is `undefined`

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
0xavro: I personally like it being explicit but no strong opinion here. Maybe @zhelezkov has some input?

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
0xavro: We need the if statement here so we don't update the state when dialect becomes `undefined` on a message send.

fsher: Ah, good point, but we can do something like: 
```
setYouCanWrite(membersContainCurrentKey ?? false);
```
Seems a bit cleaner, wdyt?

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
fsher: Ok, we can come back to this discussion later on, but let's go with explicit for now then.

fsher (approved):
LGTM, one minor comment left and good to go!

packages/dialect-react/components/DialectContext/index.tsx
fsher: Even better, great job!

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
0xavro: Yeah, let's go with this.

PR #53, opened 2022-04-06
Fix typing long msg without space will expand whole chat

0xkokoro:
When you type a long msg without space for instance when you type a long URL or an address, the msg will expand the whole chat unexpectedly. Use won't be able to see the send button or other msgs.

**Before:**
https://user-images.githubusercontent.com/93460079/161997413-5ee5658e-c68a-4b55-a2ec-b777b1bf64cc.mov

**After:**
https://user-images.githubusercontent.com/93460079/161997539-fcb1fc25-d323-467f-aa14-b5fc66f25f66.mov

### DIFF
--- a/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
@@ -31,16 +31,17 @@ export default function MessageInput({
       <div className="dt-flex dt-flex-col dt-pb-2 dt-mb-2">
         <form onSubmit={onSubmit}>
           <div className="dt-relative">
-            <div className="dt-text-sm dt-break-words dt-py-1 dt-pl-2 dt-pr-11">
+            <div style={{ width: '24rem'}} className="dt-text-sm dt-break-words dt-py-1 dt-pl-2 dt-pr-11">
               {text || 'h'}
             </div>
             <div className="dt-absolute dt-top-0 dt-w-full dt-h-full dt-flex dt-flex-grow dt-items-center">
               <Textarea
+                style={{ width: '29rem'}}
                 value={text}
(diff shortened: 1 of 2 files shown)


packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
```diff
@@ -31,16 +31,17 @@ export default function MessageInput({
       <div className="dt-flex dt-flex-col dt-pb-2 dt-mb-2">
         <form onSubmit={onSubmit}>
           <div className="dt-relative">
-            <div className="dt-text-sm dt-break-words dt-py-1 dt-pl-2 dt-pr-11">
+            <div style={{ width: '24rem'}} className="dt-text-sm dt-break-words dt-py-1 dt-pl-2 dt-pr-11">
               {text || 'h'}
             </div>
             <div className="dt-absolute dt-top-0 dt-w-full dt-h-full dt-flex dt-flex-grow dt-items-center">
               <Textarea
+                style={{ width: '29rem'}}
```
0xkokoro, line 39: This is the default chat width of the parent container but for some reason it got overwritten here.

zhelezkov (changes requested):
fixed width doesn't seems as a good fix, since it prevents text area resizing in inbox mode
Please see attached screenshot 
Screenshot 2022-04-06 at 18 56 01

Kiryous (changes requested):
Please find a solution without fixed width; It seems there is a need to do some magic with flex-grow of messaging column

cbosborn:
@0xkokoro thanks for working on this. do we want to close this PR or find a solution for it?

Kiryous:
btw, guys! I gonna pick up this one, found a simple solution

Kiryous:
Fixed in https://github.com/dialectlabs/react/pull/84

PR #54, opened 2022-04-07
Fix shrinking textbox with spaces

0xavro:
Noticed the text area shrinking when there are only spaces

Before:
textbox-before

After:
textbox-after

### DIFF
--- a/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
@@ -40,7 +40,7 @@ export default function MessageInput({
         <form onSubmit={onSubmit}>
           <div className="dt-relative">
             <div className="dt-text-sm dt-break-words dt-py-1 dt-pl-2 dt-pr-11">
-              {text || 'h'}
+              {text.trim() || 'h'}
             </div>
             <div className="dt-absolute dt-top-0 dt-w-full dt-h-full dt-flex dt-flex-grow dt-items-center">
               <Textarea


packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
```diff
@@ -40,7 +40,7 @@ export default function MessageInput({
         <form onSubmit={onSubmit}>
           <div className="dt-relative">
             <div className="dt-text-sm dt-break-words dt-py-1 dt-pl-2 dt-pr-11">
-              {text || 'h'}
+              {text.trim() || 'h'}
```
zhelezkov, line 43: I think it might be better to have text trimmed in state, but not only in visualization. otherwise it might create misleading to user. e.g. it shows empty input box, but in reality it contains symbols that got trimmed

zhelezkov (changes requested):

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/MessageInput.tsx
0xavro: The text gets rendered in the `<Textarea ... />` below on line 47. So what the user sees is what they send in the protocol. I believe this is just a placeholder to ensure the div doesn't collapse.

zhelezkov: ok, got it

zhelezkov (approved):
lgtm
I just think if the overall layout is fine(I mean absolutely position text area with this hack), but I think it might be out of scope for this task.

@fsher wdyt?

PR #58, opened 2022-04-08
Add Glow as a wallet option

vpontis:
There are still some issues that I'm looking into that are caused by type mismatch with the wallet adapter upgrade.


zhelezkov:
Hey @vpontis!

This is looks good, I just want to do the testing and include glow in other examples.
I've tried to run it on devnet, but I get this error when trying to sign a transaction
`Glow detected an issue with the requested transaction and it will likely fail`. 
Do you think I'm missing something?

And can you advice how could I switch glow wallet to localnet?

Thanks in advance!

vpontis:
Hi @zhelezkov — you can't use Glow on localnet since we don't make RPC requests directly from the wallet, we do it from our backend. You can use it in Mainnet or Devnet.

We also require the web client to specify the network rather than defaulting to whatever network the user selected in the wallet. This is for security reasons — I think the way Phantom does it right now (default to network chosen in wallet) is really insecure.

zhelezkov:
@vpontis Oh, got it, thanks for your reply, it helped a lot!

Could you please suggest, how web client could specify the network? I've been trying to find it as a parameter in wallet-adapter, but no luck

zhelezkov:
@vpontis Oh, got it, thanks for your reply, it helped a lot!

Could you please suggest, how web client could specify the network? I've been trying to find it as a parameter in wallet-adapter, but no luck

vpontis:
Hi @zhelezkov I made this :) https://github.com/luma-team/glow-js

zhelezkov:
cool, thanks
we'll take a look

Meanwhile glow support is added to dialect examples and successfully tested, so closing this pr :)

PR #59, opened 2022-04-09
fix: missing borders

Kiryous:
Related to https://github.com/tailwindlabs/tailwindcss/issues/938

If the project is not using the Tailwind lib, they don't have the following preflight borders: 
```css
*,
::before,
::after {
border: 0px solid #e5e7eb;
}
```

Therefore borders are not showing if border-style is not set explicitly.

We also could not use the prefixed preflight since it will override every class because `.dialect *` more specific than `.dt-border-b`
```css
/* Couldn't do so*/
.dialect *,
.dialect ::before,
.dialect ::after {
border: 0px solid #e5e7eb;
}
``` 

So the solution is to set the default border style only for desired side

### DIFF
--- a/packages/dialect-react-ui/index.css
+++ b/packages/dialect-react-ui/index.css
@@ -50,7 +50,7 @@
   border-left-style: solid;
 }
 
-@media (min-width: 640px) {
+@media screen(sm) {
   .sm\:dt-border {
     border-style: solid;
   }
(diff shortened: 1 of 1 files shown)


zhelezkov (approved):
looks good

maybe a dumb question, but looks like we hardcode breakpoints. Do you think we can avoid it somehow?

PR #62, opened 2022-04-09
Display twitter handle for verified wallet address using Cardinal

0xkokoro:
**Inbox**
Screen Shot 2022-04-13 at 3 17 30 PM

**Chat**
Screen Shot 2022-04-13 at 3 18 53 PM
Screen Shot 2022-04-13 at 3 18 44 PM


examples/inbox/package.json
```diff
@@ -11,6 +11,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.16.7",
+    "@cardinal/namespaces-components": "2.5.0",
```
fsher, line 14: MInor: I'd change this to ^2.5.0. so that any other minor version (>=2.5.0) can also used, so that integrating repos, that might use cardinal won't have to install 2 versions of the same package. Wdyt?

packages/dialect-react/components/DialectContext/index.tsx
```diff
@@ -473,7 +474,9 @@ export const DialectProvider = (props: PropsType): JSX.Element => {
 
   return (
     <DialectContext.Provider value={value}>
-      {props.children}
+      <WalletIdentityProvider>
```
fsher, line 477: Do we need cardinal in react package, since it's only a visual thing at the moment?

I would move this provider to NotificationsButton and ChatButton where all providers are currently located. That way we localize those changes only to react-ui, since it's only necessary for UI part of dialect at the moment, but we can improve on this later. Wdyt?

fsher (changes requested):
Well done! Minor point about versioning and a question about the provider, please take a look.

examples/inbox/package.json
0xkokoro: yea that makes perfect sense to me. just updated. btw do we have to do the same for `packages/dialect-react-ui/package.json`

packages/dialect-react/components/DialectContext/index.tsx
0xkokoro: yeaa this also makes sense to me updated so that we are only inject cardinal context in the UI components

packages/dialect-react-ui/components/Chat/screens/Main/ThreadsList/MessagePreview.tsx
```diff
@@ -64,7 +60,16 @@ export default function MessagePreview({
       </div>
       <div className="dt-flex dt-grow dt-border-b dt-border-neutral-600 dt-justify-between dt-truncate dt-pr-2">
         <div className="dt-flex dt-flex-col dt-max-w-full dt-truncate">
-          {dialect?.dialect.members.length > 0 && <div>{otherMemberStr}</div>}
+          {dialect?.dialect.members.length > 0 &&
+            (program?.provider.connection ? (
+              <CardinalDisplayAddress
+                connection={program?.provider.connection}
+                publicKey={otherMembers[0].publicKey}
+                showTwitterIcon={true}
```
Kiryous, line 68: I guess, here should be some condition. Right no
Screenshot 2022-04-14 at 11 40 43
w the twitter icon showed even if there is no cardinal identity

Kiryous (changes requested):
Nice job! Check the comment about handling addresses without cardinal

packages/dialect-react-ui/components/CardinalAddress/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  );
+};
+
+const DisplayAddressNew = ({
+  connection,
+  address,
+  height = "13",
+  width = "300",
+  style,
+  color = "#FFFFFF",
```
Kiryous, line 45: Let's use pass className and use tailwind styles here, like `className="dt-flex dt-gap-1.5"`, etc

packages/dialect-react-ui/components/CardinalAddress/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const { displayName, loadingName } = useAddressName(connection, address);
+
+  if (!address) return <></>;
+  return loadingName ? (
+    <div
+      style={{
+        ...style,
+        height,
+        width,
+        overflow: "hidden",
```
Kiryous, line 63: minor: let's run the prettier against new files using the .prettierrc from the root

Kiryous (changes requested):
Could you convert styles to tailwind and run prettier, please?

packages/dialect-react-ui/components/Chat/screens/Main/ThreadsList/MessagePreview.tsx
0xkokoro: hey nice catch! added conditional rendering for this

packages/dialect-react-ui/components/CardinalAddress/index.tsx
0xkokoro: changed the styling to use tailwind

packages/dialect-react-ui/components/CardinalAddress/index.tsx
0xkokoro: ran prettier on the dialect-react-ui/*

cbosborn (approved):
lgtm! nice job @0xkokoro & @0xavro

fsher (approved):
lgtm, for history, some tech debt tasks were created and will be addressed later on

PR #63, opened 2022-04-09
add dev command to packages which runs all builds in watch mode

zhelezkov:


fsher (approved):
Great addition, thanks! Well done.

PR #64, opened 2022-04-11
WIP esm build fix

zhelezkov:
this pr contains:
- new tsconfigs, which extends each other, so no copy pasting, settings revisited, and it provides correct esm/cjs builds
- updated wallet adapter, cause newer builds has convenient esm/cjs builds
- fixed examples, clean up next config, since transpiling not required anymore
- added starters to verify dev&build modes works:
  - cra
  - nextjs
- convenient dev modes

todo:
- [x] validate&update sollet integration, make sure that encryption is not broken, since wallet adapter has a bit new api
- [x] cleanup starters(better ui styles)
- [x] update readme, describe how to start locally
- [x] update website (https://github.com/dialectlabs/web/pull/15)


packages/dialect-react/tsconfig.json
```diff
@@ -1,34 +1,13 @@
...
   "include": [
     "**/*.ts",
     "**/*.tsx"
   ],
   "exclude": [
-    "node_modules",
-    "lib"
-  ],
-  "type": "module"
+    "node_modules"
```
fsher, line 11: I would still exclude lib folder, to be not included in the compilation, unless there is a specific reason why you've removed it.

tsconfig.json
```diff
@@ -1,33 +1,39 @@
...
-  "exclude": [
-    "node_modules",
-    "styles"
-  ],
-  "type": "module"
+    "declaration": true,
+    "declarationMap": true,
+    "inlineSources": true
+    // other
+    // "allowJs": true,
```
fsher, line 37: Let's remove this line, since unused.

packages/dialect-react/components/ApiContext/index.tsx
```diff
@@ -36,27 +36,25 @@ type PropsType = {
...
 
+export type { WalletName };
 export type WalletType = WalletContextState | AnchorWallet | null | undefined;
 export type ProgramType = anchor.Program | null;
 
-export const WalletName = { ...AdapterWalletName, Anchor: 'Anchor' as const };
-export type WalletNameType = keyof typeof WalletName;
-
-export const getWalletName = (wallet: WalletType): WalletNameType | null => {
+export const getWalletName = (wallet: WalletType): WalletName | null => {
```
fsher, line 43: I'm guessing you've removed the WalletName enum/object in order to avoid the require of the `wallet-adapter-base`. 

If it's not much to ask, but can you, please implement some enum/object that matches that object so that non-ts project can still avoid using "magic strings"? Or is there a way to still use the WalletName enum from adapter base to implement the object?

packages/dialect-react/components/ApiContext/index.tsx
```diff
@@ -36,27 +36,25 @@ type PropsType = {
   dapp?: string; // base58 public key format
 };
 
+export type { WalletName };
```
fsher, line 39: Also, does WalletName contain `Anchor` as well? I see that the casting went through, which is great.

fsher (approved):
Amazing job, thank you @zhelezkov! 

Approving in advance, most things look great. Left some comments related to the WalletNames, please check them out.

packages/dialect-react/tsconfig.json
zhelezkov: good point. I don't think that we'll have ts files in lib, but yeah, why not

packages/dialect-react/components/ApiContext/index.tsx
zhelezkov: no, not really, that's what has been changed in wallet adapter public api, they removed this enum..
but I'll do my best to avoid magic strings

packages/dialect-react/components/ApiContext/index.tsx
zhelezkov: it's just a bit different in a new version, there's no such enum at all. I guess they did it in order to avoid compile time dep. So I created an alternative enum in our package

examples/chat/components/Wallet/index.tsx
```diff
@@ -5,15 +5,16 @@ import {
...
+  PhantomWalletAdapter,
+  SolletWalletAdapter,
 } from '@solana/wallet-adapter-wallets';
 import {
   WalletModalProvider,
   WalletMultiButton,
 } from '@solana/wallet-adapter-react-ui';
 import { clusterApiUrl } from '@solana/web3.js';
 import { PhantomIcon } from '../Icon';
+import { GlowWalletAdapter } from '@solana/wallet-adapter-wallets/lib/cjs';
```
fsher, line 17: Is the `lib/cjs` import intentional?

packages/dialect-react/utils/wallets.ts
```diff
@@ -0,0 +1,4 @@
+export enum Wallets {
```
fsher, line 1: Sorry to bring this up again, but can we add more wallets here? We use Phantom, Glow and others. Ideally this should extend the WalletName since it's exactly what we want.

Also does it make sense to update the getWalletName function as well?

And lastly, I think it's conventional for enum names to be singular. To avoid collision we can either name it `SupportedWallet` or (if we don't care, as it was done initially) - `WalletName`. wdyt?

fsher (commented):
Thanks for resolving my comments! Left additional comments behind, sorry. Please take a look!

packages/dialect-react/components/ApiContext/index.tsx
fsher: I would use the enum you've introduced as the type for this, wdyt?

examples/chat/components/Wallet/index.tsx
zhelezkov: no, my bad

PR #70, opened 2022-04-14
Fix chat styles (scrollbars, paddings, avatart margin) & add links

Kiryous:
## Context
It's actually a OS-specific setting how to show scrollbars when not scrolling. Defaults to "only while scrolling" on macOS, and to "always" on windows, linux.
Screenshot 2022-04-14 at 13 57 53

These PR style scrollbars to make them more elegant if shown (`scrollbar` key introduced in theme variables). It also hides scrollbars from textarea completly, since textarrea resizes with new text added.

### before:
Screenshot 2022-04-14 at 13 32 49

### after:
Screenshot 2022-04-14 at 13 37 43

### DIFF
--- a/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
@@ -127,7 +127,7 @@ export default function Thread() {
               <div
                 className={cs(
                   otherMessageBubble,
-                  'dt-max-w-xs dt-flex-row dt-flex-shrink'
+                  'dt-max-w-xs dt-flex-row dt-flex-shrink dt-ml-1'
                 )}
               >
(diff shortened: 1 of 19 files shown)


packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/Thread.tsx
```diff
@@ -97,12 +127,35 @@ export default function Thread() {
...
+                        decoratedText: string,
+                        key: number
+                      ) => (
+                        <A
+                          target="blank"
+                          className={textStyles.link}
+                          href={decoratedHref}
+                          key={key}
+                        >
+                          {decoratedText.length > 32
```
fsher, line 151: I'd recommend moving "magic numbers" (in this case 32) to a constant.

fsher (approved):
Great job, sir! Very well done. Left a minor comment, please address

PR #71, opened 2022-04-15
when new wallet data recieved reset dialect address

AlekseiPetrovDev:
Initial bug: route away from encrypted message if switching to phantom

When wallet was changed, have to reset address dialect to rest not valid thread anymore.

Some videos with current behaviour are attached.. 

https://user-images.githubusercontent.com/78972087/163554494-87fc920d-97da-4789-95fb-2fd1fe49b92e.mov

https://user-images.githubusercontent.com/78972087/163554506-d3607325-77c5-484f-ab2a-5382f2ce2098.mov

### DIFF
--- a/packages/dialect-react-ui/CHANGELOG.md
+++ b/packages/dialect-react-ui/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [UNRELEASED]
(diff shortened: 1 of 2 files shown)


fsher (approved):
LGTM!

Kiryous (approved):
Looks good!

PR #72, opened 2022-04-15
add hash to message array to calc difference between message arrays

AlekseiPetrovDev:
When two threads has the same amount of message react can't detect switching between threads. 
Suggest to create hash based or array data. Worries performance on quite big arrays could be improved somehow in future.


packages/dialect-react/components/DialectContext/index.tsx
```diff
@@ -274,10 +274,13 @@ export const DialectProvider = (props: PropsType): JSX.Element => {
     const hasNewMessage =
       wallet &&
       dialect?.dialect &&
-      messages.length !== dialect.dialect.messages.length;
+      (messages.length !== dialect.dialect.messages.length 
+        // Could be there performacne issue to calc hash for long array?
+        || getMessageHash(dialect.dialect.messages) !== getMessageHash(messages));
```
fsher, line 279: If it's a heavy operation to hash (even though I think that sha1 shouldn't be too heavy, we don't have arrays of astronomical size), feel free to useMemo it and use that.

Also, I think that we don't need to the check the length any more since hashing will prove the change. wdyt?

fsher (approved):
Well done, sir! Please address the comments and good to go.

packages/dialect-react/components/DialectContext/index.tsx
AlekseiPetrovDev: Yes, message.length could be removed in this case. Will updated PR

Kiryous: Maybe a silly question, but why we introduced the condition here to update messages only if lenght is not matched, the fact `setMessages` is called in useEffect with `wallet, dialect?.dialect` dependencies should already do the trick, no?

@fsher

PR #73, opened 2022-04-15
fix light theme text for chat

AlekseiPetrovDev:


packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/index.tsx
```diff
@@ -89,7 +89,7 @@ const ThreadPage = ({
           <icons.back />
         </div>
         <div className="dt-flex dt-flex-col dt-items-center">
-          <span className="dt-text-base dt-font-medium dt-text-white">
+          <span className="dt-text-base dt-font-medium dt-text">
```
fsher, line 92: do we need to leave this class at all?

I would suggest even to take font styles from theme and add them here (but may be left out).

fsher (approved):
LGTM, minor comment but good to go in general! Thanks!

PR #76, opened 2022-04-18
fix: use preflighted <img /> tag with max-height already set

Kiryous:
## Before
Screenshot 2022-04-18 at 20 46 41

## After
Screenshot 2022-04-18 at 20 46 12


0xavro (approved):
Looks good! Thanks for fixing

PR #78, opened 2022-04-19
Support encryption while using @saberhq/use-solana as wallet provider

zhelezkov:
also added new starter to test integration with use-solana package


packages/dialect-react/utils/helpers.ts
```diff
@@ -26,3 +27,22 @@ export const isWalletContextState = (
...
+  if (!wallet) {
+    return null;
+  }
+  const isSaberUseSolana = wallet.constructor.name === 'WrappedWalletAdapter';
+  if (isSaberUseSolana) {
+    return wallet.adapter.adapter;
+  }
+
+  return wallet.adapter;
+};
```
zhelezkov, line 48: this like the main crucial part

Kiryous (approved):
LGMT!

PR #79, opened 2022-04-19
Refactor display address and avatar to include cardinal integration 

0xkokoro:


packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/index.tsx
```diff
@@ -94,7 +82,15 @@ const ThreadPage = ({
         </div>
         <div className="dt-flex dt-flex-col dt-items-center">
           <span className="dt-text-base dt-font-medium dt-text-white">
-            {dialect ? displayAddress : 'Loading...'}
+            {dialect ? (
```
fsher, line 85: Does it make sense to include the loading state into DisplayAddress as well?

fsher (approved):
LGTM! Left a minor comment there, check it out!

packages/dialect-react-ui/components/Chat/screens/Main/pages/ThreadPage/index.tsx
0xkokoro: Thanks for the review! 🤔 I would assume non since we don't include any loading for MessagePreview. Happy to put discuss about this in more detail later.

PR #80, opened 2022-04-20
increase token life up to 3h + save token into session storage

AlekseiPetrovDev:


packages/dialect-react/api/web2.ts
```diff
@@ -30,6 +30,56 @@ const signPayload = async (wallet: WalletContextState, payload: Uint8Array) => {
...
+  );
+
+  return {
+    expirationTime: expirationTime.toString(),
+    base64Signature
+  };
+}
+
+const saveToken = (token: TokenType) => {
+  window.sessionStorage.setItem("expirationTime", token.expirationTime.toString());
```
fsher, line 58: Just realized that this is going to break in cases where a project is using Nextjs SSR. Let's wrap with a small check that `window` exists or similar. wdyt?

fsher: Also, quick question, why not wrap these 2 entries into one, assuming they are paired together.

fsher (changes requested):
Well done! Left some comments, probably needs to be done beforehand. Please check them out!

packages/dialect-react/components/ApiContext/index.tsx
```diff
@@ -109,6 +110,7 @@ export const ApiProvider = ({ dapp, children }: PropsType): JSX.Element => {
 
   useEffect(() => {
     if (isWalletConnected) {
+      removeToken();
```
fsher, line 113: just out of curiosity, why not remove the token when the wallet gets disconnected rather than when connected?

fsher (approved):
LGTM, please check last comment

packages/dialect-react/components/ApiContext/index.tsx
AlekseiPetrovDev: make sense. But also anyway if wallet changed isWalletConnected will be trigger and old token need to be removed. But I think we can move this line above this condition

AlekseiPetrovDev: @fsher  what do you think?

cbosborn (approved):

PR #81, opened 2022-04-20
Fix/redirect existing thread

Kiryous:
If the thread with wallet address entered already present in dialects — redirect to it.


zhelezkov (approved):
lgtm!

PR #84, opened 2022-04-21
fix: prevent long message from expanding entire chat

Kiryous:
### Before:
Screenshot 2022-04-21 at 21 06 59

### After:
Screenshot 2022-04-21 at 21 09 11

Explanation: default min-width for flex elements is auto, overcompeting width and max-width, which results in overflowing


0xkokoro (approved):
🚢  it! Thanks for fixing this :)

PR #85, opened 2022-04-21
fix: opening new thread right after creation

Kiryous:


cbosborn:
@Kiryous nice catch. I'm nervous about using timeouts to solve synchronicity issues though. Can we do this deterministically with no timeouts?

fsher:
It's always fun to see fixes related to the event loop behavior, especially with tasks/microtasks :D
But I have to agree with @cbosborn on this. Do you have an explanation why this is happening and why setTimeout is the right fix here, @Kiryous. Nice catch tho!

fsher:
Any thoughts on this @Kiryous?

Kiryous:
@fsher finally had time to look into it, fixed without setTimeout! Please, have a look.

PR #86, opened 2022-04-21
fix: remove fetchingError, instead use the errors provided by SWR to …

Kiryous:
…be able to recover

Before/After: https://www.loom.com/share/d007621b200a486a980969cd1be34de6


zhelezkov (approved):
LOOM EXPLANATION IS AWESOME

PR #87, opened 2022-04-21
fix: reset messages if the wallet disconnected

Kiryous:
before:
https://user-images.githubusercontent.com/2504771/164545225-5ed5decf-878a-498e-ae90-5a18a8575472.mov


zhelezkov (approved):
looks solid

PR #88, opened 2022-04-22
Chore/update exports

fsher:
- Removed wildcard imports
- Update CHANGELOG entry
- Remove Footer from Notification center


packages/dialect-react/index.ts
```diff
@@ -1,10 +1,38 @@
-export * from './components/ApiContext';
-export * from './components/DialectContext';
-export * from './api';
-export * as DialectErrors from './utils/errors';
-export { connected } from './utils/helpers';
-
+import * as DialectErrors from './utils/errors';
```
Kiryous, line 1: seems like one wildcard is still here

Kiryous (changes requested):

packages/dialect-react/index.ts
Kiryous: or this kind of named wildcard is not a problem?

fsher: This one is fine, since it's an `import` wildcard first and a named export then (at least I hope that it's fine)

fsher: I think the problem is more related to an export wildcard

Kiryous (approved):

PR #91, opened 2022-04-23
fix web2 error handling

AlekseiPetrovDev:


packages/dialect-react-ui/components/Notifications/EmailForm.tsx
```diff
@@ -49,6 +46,10 @@ export function EmailForm() {
   const [isEmailEditing, setEmailEditing] = useState(!emailObj?.enabled);
   const [emailError, setEmailError] = useState<ParsedErrorData | null>(null);
 
+  const [verificationCodeError, setVerificationCodeError] = useState<Error | null>(null);
```
fsher, line 49: MIght be a weird question, but why are we handling these errors separately. I believe they can't happen simultaneously. Why not store them in a normalized matter (convenient for the UI) in one useState? wdyt?

packages/dialect-react-ui/components/Notifications/SmsForm.tsx
```diff
@@ -49,6 +46,10 @@ export function SmsForm() {
   const [isSmsNumberEditing, setSmsNumberEditing] = useState(!smsObj?.enabled);
   const [smsNumberError, setSmsNumberError] = useState<ParsedErrorData | null>(null);
 
+  const [verificationCodeError, setVerificationCodeError] = useState<Error | null>(null);
```
fsher, line 49: Same here

packages/dialect-react-ui/components/Notifications/EmailForm.tsx
fsher: Ah, it was done like that before. I think we should refactor, wdyt?

fsher (changes requested):
Nice catch! I've left a bit of comments there, could you please take a look?

packages/dialect-react-ui/components/Notifications/EmailForm.tsx
Kiryous: Agreed! Additional q: why did you move the errors from ApiContext to the EmailForm itself at the first place? Since we would have SMS and telegram it make sense to keep it in one place. 

@AlekseiPetrovDev, wdyt?

Kiryous (changes requested):

AlekseiPetrovDev:
@[Kiryous](https://github.com/Kiryous) Because in that case you have APIContext (which let's say store) and 3 react component which rely on this store. Basically if error happens (not matter which error ), you will same error in 3 places. Also technically API context shoudn't story anything, it's kind of proxy which should be pretty simple: request -> response. Error handling looks like more UI part, and there we can also make some difference and specific case in future

AlekseiPetrovDev:
@Kiryous @fsher wdyt?

Kiryous:
Agreed!

Btw, one of @fsher's questions left unaddressed:
>MIght be a weird question, but why are we handling these errors separately. I believe they can't happen simultaneously. Why not store them in a normalized matter (convenient for the UI) in one useState? wdyt?

Meaning inside the EmailForm/SMSForm

AlekseiPetrovDev:
Not need

PR #92, opened 2022-04-23
Avoid fetching images urls multiple times

0xkokoro:
Before we would fetch the image url every time the avatar is displayed. So if a user sends 100 messages in the threads then we would have to fetch the same image url 100 times. This is clearly not scalable as the number users and messages grow on our platform.

**Before**
Screen Shot 2022-04-23 at 12 35 09 PM

**After**
Screen Shot 2022-04-23 at 12 37 07 PM

### DIFF
--- a/packages/dialect-react-ui/components/Avatar/index.tsx
+++ b/packages/dialect-react-ui/components/Avatar/index.tsx
@@ -3,9 +3,11 @@ import type { PublicKey } from '@project-serum/anchor';
 import { useTheme } from '../common/ThemeProvider';
 import { useApi } from '@dialectlabs/react';
 import type { Connection } from '@solana/web3.js';
-import { useAddressImage } from '@cardinal/namespaces-components';
+import { tryGetImageUrl } from '@cardinal/namespaces-components';
+import { getReverseEntry, breakName, formatName } from '@cardinal/namespaces';
 import { HiUserCircle } from 'react-icons/hi';
 import { Img } from '../common/preflighted';
+import useSWR from 'swr';
 
 const containerStyleMap = {
   regular: 'dt-w-14 dt-h-14',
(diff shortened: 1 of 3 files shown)


packages/dialect-react-ui/components/Avatar/index.tsx
```diff
@@ -33,18 +35,25 @@ const CardinalAvatar = ({
   placeholder?: React.ReactNode;
   className?: string;
 }) => {
-  const { addressImage, loadingImage } = useAddressImage(connection, address);
+  const reverseEntry = useSWR([connection, address], getReverseEntry);
```
fsher, line 38: I think we should abstract this into a custom hook inside react-ui package, wdyt?
Similar to `useAddressImage`, but our own.

fsher (changes requested):
Well done! Just one comment, I think we should abstract the hook to fetch images, otherwise good to go imo!

packages/dialect-react-ui/components/Avatar/index.tsx
cbosborn: @fsher what do you think about this being a part of the member object in the `react` package? feels agnostic to react-ui specifically, could be reused e.g. in react native, etc.

cbosborn: should clarify: I'm on board with a quick implementation for now, fwiw

packages/dialect-react-ui/components/Avatar/index.tsx
```diff
@@ -33,18 +35,25 @@ const CardinalAvatar = ({
...
+  const displayName = formatName(
+    reverseEntry.data?.parsed.namespaces,
+    reverseEntry.data?.parsed.entryName
+  );
+  const [_namespace, handle] = displayName ? breakName(displayName) : [];
+  const addressImage = useSWR(handle, tryGetImageUrl);
```
cbosborn, line 44: handle may not be specific enough, given it's global name spacing. maybe `/twitter-image-addresses/${handle}` or something

cbosborn (changes requested):
@0xkokoro nice work, just a cal smaller comments. I'm open to no serious refactors until later

packages/dialect-react-ui/components/Avatar/index.tsx
fsher: That's a good question and I think you're right. It makes sense, since react provides the data and react-ui renders it. So yeah, probably better to move to `react`. Agree.

packages/dialect-react-ui/package.json
```diff
@@ -28,6 +28,7 @@
     "lint:fix": "yarn lint --fix"
   },
   "dependencies": {
+    "@cardinal/namespaces": "^1.1.0",
     "@cardinal/namespaces-components": "2.5.0",
```
fsher, line 32: Please check if this package has newer version, I'm getting concerned with the amount of anchor packages we're installing :D

PR #96, opened 2022-04-27
fix[wip]: don't fetch web2 addresses if not enabled

Kiryous:
- [ ] fix type issue on NotificationsButton


AlekseiPetrovDev:
Have a suggestion: squash all work related commits into 1 commit and prettier make as separate comment, in that case will be possible to check changes avoid checking prettier diff. I realised I have the same issue in my PR @Kiryous @fsher what do you think?

Kiryous:
Hey! Do you mean for this PR or a general strategy? Actually if everyone is running "Format on save" with the .prettierrc config, should be always up to date

Kiryous:
@fsher hey! Could you help nail down the type error with channelOptions, please?

fsher:
Sorry for a late reply, what seems to be the problem? @Kiryous

PR #97, opened 2022-04-27
Feature/customize popup animation

fsher:


packages/dialect-react-ui/components/Chat/screens/Error.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react';
```
zhelezkov, line 1: shouldn't it cause problems after transpilation?

zhelezkov (approved):
looks solid, one q

packages/dialect-react-ui/components/Chat/screens/Error.tsx
fsher: nope, since tsconfig is set to `react-jsx`, a new import gets added: `import { jsx as _jsx } from 'react/jsx-runtime'`

zhelezkov: coool

Kiryous (approved):
Great work! Added an animation ovverride to example

PR #100, opened 2022-04-30
Expose notifications modal and add optional back buttons

0xavro:
Exposing our notifications modal and allow passing in of optional back buttons for the realms integration

component-back-button


Kiryous (approved):
Great job! Looks good, left just a few comments

PR #103, opened 2022-05-06
bundle size reducing(first iteration)

zhelezkov:
with this flag webpack will do more agressive tree shaking, so it'll help to avoid unnecessary dependencies at runtime(e.g. cardinal library which as unusued if we use notification mode)

screenshots attached, please do not look at absolute values, it's not real numbers(cause multiple libraries could be deduplicated) and further reducing can be done.

as per discussion with @cbosborn this is a first iteration with a quick(hacky?) fix to reduce bundle size. Further improvements will come soon..

As we can see current reduction is around **20%** due to cardinal exclusion

before(700kb):
Screenshot 2022-05-06 at 22 42 46

after(500kb):
Screenshot 2022-05-06 at 22 42 01

### DIFF
--- a/packages/dialect-react-ui/CHANGELOG.md
+++ b/packages/dialect-react-ui/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- mark library as side effect free
+
 ## [0.8.3] - 2022-05-08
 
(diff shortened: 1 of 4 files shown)


fsher (approved):
LGTM, awesome fix really :D

PR #107, opened 2022-05-10
fix: allow passing not only walletContext with adapter key, but also …

Kiryous:
…walletAdapter itself


zhelezkov (approved):
This is looks good, and shouldn't break the use-solana and wallet-adapter providers.

But, could you please explain it a bit? I mean, what kind of wallet adapter library tulip uses and why it's like this?
Maybe it also worth to create a starter? To simplify testing in future..

Kiryous:
@zhelezkov hey! Thanks for checking out. Sure, the thing is — some projects use custom stores (e.g. mobx, redux) to store wallet data. Inside state of such store the wallet under "wallet" key with actual provider type of PhantomWalletProvider, etc.

PR #111, opened 2022-05-19
Add solana name service support

0xavro:
Adding support for solana name service.
Currently Twitter handles take precedence over SNS and SNS takes precedence over the raw wallet public key.

Next step would be to add SWR support for fetching the SNS names.

Screen Shot 2022-05-18 at 11 48 24 PM

### DIFF
--- a/packages/dialect-react-ui/components/DisplayAddress/index.tsx
+++ b/packages/dialect-react-ui/components/DisplayAddress/index.tsx
@@ -2,6 +2,7 @@ import { useAddressName } from '@cardinal/namespaces-components';
 import { getNameEntry } from '@cardinal/namespaces';
 import { TwitterIcon } from '../Icon/Twitter';
 import type { Connection, PublicKey } from '@solana/web3.js';
+import { SolanaNameServiceName } from '../common';
 import { useApi } from '@dialectlabs/react';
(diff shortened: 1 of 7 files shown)


packages/dialect-react-ui/components/DisplayAddress/index.tsx
```diff
@@ -81,6 +82,42 @@ const TwitterHandle = ({
...
+  return `${domain.slice(0, 4)}...${domain.slice(domain.length - 2)}`;
+};
+
+const SolanaName = ({
+  address,
+  displayName,
+  loadingName,
+  dimensionClassName = '',
+}: {
+  address: PublicKey | undefined;
```
Kiryous, line 98: coulde be `address?`

Kiryous (approved):
Looks great! Just one q in comments and also: should this PR add resolving domains on the CreateThread view?

packages/dialect-react-ui/components/DisplayAddress/index.tsx
0xavro: yep let's do this instead

cbosborn (approved):
lgtm. let's get thread creation support in a follow up pr

PR #118, opened 2022-05-30
fix: update README to include an inbox example, ordering inbox above chat

kevindingens:
https://www.notion.so/change-default-examples-in-react-README-3b49189a9b3d4128851a30df07396a79

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,31 @@ The component above is a self-contained button that opens a notifications modal 
 2. `publicKey` – The public key associated with your project's messaging keypair. All notifications sent to your users are signed and written on-chain using this keypair.
 3. `network` – Which network to target. `localnet`, `devnet`, & `mainnet-beta` are supported.
 
+### Embed a full inbox view in your website
+
+```typescript
+import '@dialectlabs/react-ui/index.css';
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,6 +64,31 @@ The component above is a self-contained button that opens a notifications modal
...
+
+return <Inbox 
+  wrapperClassName="p-2 h-full overflow-hidden rounded-2xl shadow-2xl shadow-neutral-800 border border-neutral-600" 
+  wallet={wallet} 
+/>;
+
+// ...
+```
+
+The component above contains a wallet's current inbox of current chats/notifications. From this inbox, a user can create, browse, and send chats of their own. Props are outlined as follows.
```
kevindingens, line 86: From what I know playing around with inbox example, this is true, but need a check

fsher: This is correct, yes!

I'd rephrase a bit, the "send chats" part bothers me a bit, sorry :D Should we do something like: `create, browse chats, and send messages.` Wdyt?

fsher (approved):
Well done! Left a minor comment. May be wrong there, feel free to address it!

README.md
kevindingens: Agreed. Updating now

PR #122, opened 2022-06-03
feat: added read receipts functionality for notifications

AlekseiPetrovDev:
Screen Shot 2022-06-03 at 16 16 38

### DIFF
--- a/packages/dialect-react-ui/components/NotificationsButton/index.tsx
+++ b/packages/dialect-react-ui/components/NotificationsButton/index.tsx
@@ -20,6 +20,7 @@ import {
 import type { Channel } from '../common/types';
 import Notifications, { NotificationType } from '../Notifications';
 import IconButton from '../IconButton';
+import clsx from 'clsx';
 
 export type PropTypes = {
   wallet: WalletType;
@@ -116,14 +117,14 @@ function WrappedNotificationsButton(
 
   return (
     <div
-      className={cs(
+      className={clsx(
         'dt-flex dt-flex-col dt-items-end dt-relative',
         colors.primary
       )}
     >
       {hasNewMessages && (
         <div
-          className={cs(
+          className={clsx(
             'dt-absolute dt-h-3 dt-w-3 dt-z-50 dt-rounded-full',
             colors.notificationBadgeColor
           )}
@@ -131,7 +132,7 @@ function WrappedNotificationsButton(
       )}
       <IconButton
         ref={bellRef}
-        className={cs(
+        className={clsx(
           'dt-flex dt-items-center dt-justify-center dt-rounded-full focus:dt-outline-none dt-shadow-md',
           colors.bg,
           bellButton
--- a/packages/dialect-react-ui/components/NotificationsButton/index.tsx
+++ b/packages/dialect-react-ui/components/NotificationsButton/index.tsx
@@ -5,6 +5,7 @@ import {
   connected,
   useApi,
   DialectProvider,
+  useDialect,
 } from '@dialectlabs/react';
 import type { WalletType } from '@dialectlabs/react';
 import { Transition } from '@headlessui/react';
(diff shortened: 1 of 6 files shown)


AlekseiPetrovDev:
This is temporary fast approach how we can achieve this. Later I would like to add Adapter wrapper and pass to adapter - connector which could be a localStorage, asynStorage or whatever based on abstract class getItem() & setItem()

packages/dialect-react-ui/components/NotificationsButton/index.tsx
```diff
@@ -69,10 +70,41 @@ function WrappedNotificationsButton(
...
   useOutsideAlerter(wrapperRef, bellRef, setOpen);
   const { setWallet, setNetwork, setRpcUrl } = useApi();
   const isWalletConnected = connected(props.wallet);
 
+  const { messages } = useDialect();
+
+  useEffect(() => {
+    if (!window) return;
+
+    const lastMsgTimestampStorage = window.localStorage.getItem('lastMsg');
```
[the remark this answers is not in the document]
fsher: Also, I think we need to also abstract the logic of getting the `lastMsg` timestamp, so that later we can just change it to server fetching?

Also-also, given that we have one localStorage, but potentially many windows related to dialect, would it make sense to track which thread this `lastMsg` is related to? What I'm saying, I think it's worth changing the storage to a map:
```
{
  [threadId: string]: string; // timestamp
}
```
That way if we decide to scale to multi-thread tracking, it would be easy to add this to our localStorage.

And final also-also-also, I think we need more control over localStorage in case of any bugs or structure changes, since we're essentially using it as our DB. Please consider adding migrations functionality to the structure (e.g. https://github.com/ragnarstolsmark/localstorage-migrator), so that we can update the structure of the localstorage to our needs, in case we update it in any way. This will give us control over something that we have no access to. Please feel free to challenge this.

packages/dialect-react-ui/components/NotificationsButton/index.tsx
```diff
@@ -103,6 +135,14 @@ function WrappedNotificationsButton(
         colors.primary
       )}
     >
+      {hasNewMessages && (
+        <div
+          className={cs(
```
fsher, line 140: Minor: `cs` is our custom function that we've started replacing with `clsx` package. Could you please update to `clsx` instead?

packages/dialect-react-ui/components/NotificationsButton/index.tsx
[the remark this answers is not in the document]
AlekseiPetrovDev: This make perfect sense, fix it, please review it

packages/dialect-react/components/ApiContext/index.tsx
```diff
@@ -221,6 +226,43 @@ export const ApiProvider = ({ dapp, children }: PropsType): JSX.Element => {
     [dapp, isWalletConnected]
   );
 
+  const getLastReadMessageWrapper = (threadId: string) => {
```
fsher, line 229: Please wrap these functions with useCallback

packages/dialect-react/components/ApiContext/index.tsx
```diff
@@ -221,6 +226,43 @@ export const ApiProvider = ({ dapp, children }: PropsType): JSX.Element => {
...
+    threadId: string,
+    timestamp: string | undefined
+  ) => {
+    if (!window || !timestamp) return;
+
+    const dialectReadReceipts = window.localStorage.getItem(
+      'dialectReadReceipts'
+    );
+
+    let dialectReadReceiptsObj: any = {};
```
fsher, line 251: why `any`?

fsher (approved):
LGTM, minor comments left

AlekseiPetrovDev:
@fsher fixed latest comments

PR #125, opened 2022-06-10
[WIP] Feature/twitter chat experience

fsher:
Screen Shot 2022-06-10 at 19 51 31

Screen Shot 2022-06-10 at 19 51 57

Screen Shot 2022-06-10 at 19 52 15

Screen Shot 2022-06-10 at 19 52 30

### Major Changes unrelated to the feature
- Added new provider - DialectUiManagmentProvider
- Chat is now wrapped with a ChatProvider to reduce prop drilling
- Added a memory Router (preliminary implementation)
- Restructured and renamed a bit

Has some wacky moments and there are a lot last-thing changes. I would really appreciate a deep review here and help in certain moments (where TODOs are). Most of the things are done, but there are some leftovers that I will continue during the review process

- [ ] Need to fully test other experiences: inbox (partially tested), notifications, popup chat
- [ ] Fix showing transition, works partially


packages/dialect-react-ui/components/Chat/provider/index.tsx
```diff
@@ -0,0 +1,37 @@
+import { createContext, FunctionComponent, useContext } from 'react';
+
+// We are duplicating this type in order to have separation of concern: it may be different to what we have in props
+type ChatType = 'inbox' | 'popup' | 'vertical-slider';
+
+interface ChatContextValue {
+  id: string;
```
zhelezkov, line 7: just thinking, shouldn't it be more dialect specific, e.g. `dialectId` to do not confuse with id from html?

packages/dialect-react-ui/components/Chat/provider/index.tsx
```diff
@@ -0,0 +1,37 @@
+import { createContext, FunctionComponent, useContext } from 'react';
+
+// We are duplicating this type in order to have separation of concern: it may be different to what we have in props
+type ChatType = 'inbox' | 'popup' | 'vertical-slider';
```
zhelezkov, line 4: looks very hacky, probably composition would play better here, but I do not insist, since it would take a huuge amount of time. Might revisit it during further refactoring. wdyt?

packages/dialect-react-ui/components/Chat/provider/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+interface ChatProviderProps {
+  id: string;
+  type: ChatType;
+  onChatClose?: () => void;
+  onChatOpen?: () => void;
+}
+
+export const ChatContext = createContext<ChatContextValue | null>(null);
```
zhelezkov, line 20: i guess it should be smth like
```suggestion
export const ChatContext = createContext<ChatContextValue>({} as ChatContextValue);
```

packages/dialect-react-ui/components/Chat/screens/CreateThreadPage/CreateThread.tsx
```diff
@@ -300,14 +301,18 @@ export default function CreateThread({
...
+  const { ui } = useDialectUiId(id);
   const { program, network, wallet, walletName } = useApi();
   const connection = program?.provider.connection;
   const { balance } = useBalance();
   const { colors, outlinedInput, textStyles, icons } = useTheme();
 
-  const [address, setAddress] = useState('');
-  const [actualAddress, setActualAddress] =
-    useState<anchor.web3.PublicKey | null>(null);
+  const [address, setAddress] = useState(receiver ?? '');
```
zhelezkov, line 314: shouldn't it be null too?

packages/dialect-react-ui/components/Chat/screens/Main/index.tsx
```diff
@@ -1,77 +1,92 @@
...
     <div className="dt-h-full dt-flex dt-flex-1 dt-justify-between dt-w-full">
       <div
         className={clsx(
           'dt-flex dt-flex-1 dt-flex-col dt-border-neutral-600 dt-overflow-hidden dt-w-full',
           {
             'md:dt-max-w-xs md:dt-border-r md:dt-flex': inbox,
-            'dt-hidden': dialectAddress || newThreadOpen,
+            'dt-hidden':
+              dialectAddress ||
+              current?.sub?.name === MainRouteName.CreateThread, // TODO: ideally we should control this with routes and this causes a visual bug with route swaps
```
zhelezkov, line 31: just wonder, what kind of visual bug?

packages/dialect-react-ui/components/Header/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+        )}
+        onClick={onHeaderClick}
+      >
+        {children}
+      </header>
+    </HeaderContext.Provider>
+  );
+};
+
+Header.Icon = function HeaderIcon({
```
zhelezkov, line 77: just wonder, why not using an arrow function?

zhelezkov (changes requested):
So many code, so many contexts..
a bit complicated, I think I'll review it a bit later one more time, need to think about new contexts and router :D

I haven't tested it yet btw, will check later!

packages/dialect-react-ui/components/Chat/provider/index.tsx
fsher: agree! Great point.

packages/dialect-react-ui/components/Chat/screens/CreateThreadPage/CreateThread.tsx
fsher: Sure!

packages/dialect-react-ui/components/Chat/screens/Main/index.tsx
fsher: Router provider is setting the child `activeRoute` in `useEffect`, so route on lower levels happens slower than this class update. That causes a quick appearance of the menu, while we are still on the `CreateThread` route.

packages/dialect-react-ui/components/common/providers/DialectThemeProvider.tsx
```diff
@@ -399,7 +421,7 @@ type PropsType = {
   children: JSX.Element;
 };
 
-export const ThemeProvider = ({
+export const DialectThemeProvider = ({
```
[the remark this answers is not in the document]
fsher: Nope, ThemeProvider is an internal thing. Mb only for Inbox (not sure).

Even then, this PR still has breaking changes with DialectUiManagementProvider.

packages/dialect-react-ui/components/Chat/screens/Main/index.tsx
zhelezkov: have you tried useLayoutEffect?

packages/dialect-react-ui/components/Header/index.tsx
zhelezkov: cool

packages/dialect-react-ui/components/Chat/screens/Main/index.tsx
fsher: useLayoutEffect will cause problems (afaik) with SSR projects, at least I remember a bunch of warnings at the time.

But I'll give it a shot, thanks for the suggestion

packages/dialect-react-ui/components/Chat/provider/index.tsx
zhelezkov: I just thought that `type` might be a bit hacky, and too overwhelming the component itself
By composition I meant to have smth like `ChatInbox`, `ChatPopup`, `ChatVerticalSlider`, which would have composed internally with reusable components

packages/dialect-react-ui/components/Chat/screens/Main/index.tsx
zhelezkov: oh I see, then should be smth like this https://github.com/streamich/react-use/blob/master/src/useIsomorphicLayoutEffect.ts#L4
to support ssr projects

packages/dialect-react-ui/components/Chat/provider/index.tsx
fsher: Great point, thanks! I suggest to discuss/plan this during next refactor of react-ui? Wdyt?

packages/dialect-react-ui/components/Chat/screens/Main/index.tsx
fsher: Yes, good point! 

Btw, useLayoutEffect is working! Nextjs isn't sending any warnings and actually it makes sense, I believe, since we don't SSR render this page on the backend

zhelezkov: cool, but anyway I think it should be isomorphic, since some of our partners could use ssr ;)

packages/dialect-react-ui/components/Chat/provider/index.tsx
zhelezkov: yes def

packages/dialect-react-ui/components/Chat/screens/CreateThreadPage/CreateThread.tsx
Kiryous: let's update this minor thing?

packages/dialect-react-ui/components/Chat/navigation.ts
```diff
@@ -0,0 +1,37 @@
+import { MainRouteName, RouteName, ThreadRouteName } from './constants';
+import type { RouterContextValue } from '../common/providers/Router';
+
+type NavigateFn = RouterContextValue['navigate'] | undefined;
```
zhelezkov, line 4: just wonder, why undefined?

zhelezkov (approved):
this is good, huuge amount of work done, one q

packages/dialect-react-ui/components/Chat/screens/CreateThreadPage/CreateThread.tsx
fsher: yes, thanks, forgot about this

packages/dialect-react-ui/components/Chat/navigation.ts
fsher: Not all windows have navigations at the moment, which means not all have `navigate`.

PR #129, opened 2022-06-15
fix: show solana name service domain without middle truncating, only …

Kiryous:
…truncate in message preview if overflows free space

### DIFF
--- a/packages/dialect-react-ui/components/DisplayAddress/index.tsx
+++ b/packages/dialect-react-ui/components/DisplayAddress/index.tsx
@@ -83,13 +83,6 @@ const TwitterHandle = ({
   );
 };
 
-const formatSolanaDomain = (domain: string): string => {
-  if (domain.length <= 6) {
-    return domain;
-  }
-  return `${domain.slice(0, 4)}...${domain.slice(domain.length - 2)}`;
-};
-
 const SolanaName = ({
   address,
   displayName,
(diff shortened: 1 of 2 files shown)


packages/dialect-react-ui/components/DisplayAddress/index.tsx
```diff
@@ -164,10 +153,12 @@ export function DisplayAddress({
   const loadingName = !displayName && !error;
   const showTwitterIcon = displayName?.includes('@');
 
-  const { data } = useSWR(
-    isMemberExist ? [connection, publicKey.toString(), 'sns'] : null,
-    fetchSolanaNameServiceName
-  );
+  // const { data } = useSWR(
```
fsher, line 156: Please revert uncomment this :)

packages/dialect-react-ui/components/DisplayAddress/index.tsx
```diff
@@ -112,11 +105,7 @@ const SolanaName = ({
       </div>
     );
   }
-  return (
-    <div className="dt-flex dt-gap-1.5">
-      {`${formatSolanaDomain(displayName)}.sol ◎`}
-    </div>
-  );
+  return <div className="dt-truncate dt-mr-0.5">{`${displayName}.sol ◎`}</div>;
```
fsher, line 108: why only mr? Previously it was `gap` (even thought I think since it had only 1 child it didn't properly work).

fsher (approved):
LGTM, minor comment

packages/dialect-react-ui/components/DisplayAddress/index.tsx
Kiryous: ahahah, yeah!

packages/dialect-react-ui/components/DisplayAddress/index.tsx
fsher: Yes, I understand. But why the mr-0.5 then? Shouldn't we adjust the margin in the parent rather than here?

PR #131, opened 2022-06-21
Fix/thread related bugs

fsher:


packages/dialect-react-ui/components/Chat/screens/ThreadPage/Settings.tsx
```diff
@@ -69,6 +72,7 @@ const Settings = ({ onCloseRequest }: SettingsProps) => {
               // TODO: properly wait for the deletion
               onCloseRequest?.();
               setDialectAddress('');
+              navigate(RouteName.Main, { sub: { name: MainRouteName.Thread } });
```
Kiryous, line 75: We could use navigation shortcuts from `components/Chat/navigation.ts` here like `showMain` to encapsulate the actual navigation considering we might change router in near future. What do you think?

Kiryous (approved):
Looks pretty straightforward and legit!

Left a conceptual question though

packages/dialect-react-ui/components/Chat/screens/ThreadPage/Settings.tsx
fsher: Sure, let's start updating to this!

PR #135, opened 2022-06-27
fix #133 wallet disconnection error

### LINKED ISSUE #133: If we are disconnecting wallet the skd not initialized error is raised
Screenshot 2022-06-27 at 17 02 51

Screencast: https://www.loom.com/share/1e3b525b733e475686554c38e5a2e0ad

zhelezkov:
fixes #133


packages/dialect-react-ui/components/Chat/index.tsx
```diff
@@ -1,69 +1,45 @@
-import { useEffect } from 'react';
+import React from 'react';
```
fsher, line 1: You don't need the full `React` import, `jsx-runtime` gets added in compile time.

packages/dialect-react-ui/components/Chat/index.tsx
```diff
@@ -75,87 +51,119 @@ function InnerChat({
...
+    },
+  } = useDialectConnectionInfo();
+
+  const { connected: isWalletConnected } = useDialectWallet();
+
+  const someBackendConnected =
+    (isSolanaShouldConnect && isSolanaConnected) ||
+    (isDialectCloudShouldConnect && isDialectCloudConnected);
+
+  const hasError = !isWalletConnected || !someBackendConnected;
```
fsher, line 123: Just realizing this now, but would be more convenient to have this aggregated information in DialectProvider, so that when developing a UI, a developer won't have to think about what exactly is wrong, and/or which backend is causing this issue. Wdyt?

No need to adjust this now, just a open question.

fsher (approved):
LGTM! Left a bit of comments there.

packages/dialect-react-ui/components/Chat/index.tsx
zhelezkov: good catch!

packages/dialect-react-ui/components/Chat/index.tsx
zhelezkov: Hm, I'm not 100% about it. Cause we already have all necessary information, and this is very app specific i thinnk..

PR #139, opened 2022-07-06
Fix: add default channels if props channel not exist

AlekseiPetrovDev:


packages/dialect-react-ui/components/NotificationsButton/index.tsx
```diff
@@ -98,7 +97,7 @@ function WrappedNotificationsButton(props: PropTypes): JSX.Element {
           // style={{ backdropFilter: 'blur(132px)' }}
         >
           <Notifications
-            channels={props.channels}
+            channels={props.channels || ['web3', 'email', 'sms', 'telegram']}
```
zhelezkov, line 100: maybe extract it out of component? so that we eliminate possible rerendering issue

Kiryous (approved):
Please, resolve conflicts, otherwise — good to go!

PR #141, opened 2022-07-08
hotfix: save dialect auth token per wallet when using v0 addresses api

Kiryous:
Context: While QA'ing dashboard we found what if user change wallet after generating token for v0 then the wrong token will be used, so this change fixes it adding a prefix.

It would be cool to update addresses management to new sdk, but it's not that quick.


AlekseiPetrovDev (approved):
LGTM

PR #142, opened 2022-07-08
fix: add text limit for title 'cause it now included in message body …

Kiryous:
…for some channels, decrease message limit to 800

### DIFF
--- a/packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
+++ b/packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
@@ -32,13 +32,14 @@ function BroadcastForm({ dapp }: BroadcastFormProps) {
   const [isSending, setIsSending] = useState(false);
   const [title, setTitle] = useState('');
   const [message, setMessage] = useState('');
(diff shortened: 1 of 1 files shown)


packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
```diff
@@ -31,6 +32,10 @@ function BroadcastForm({ dapp }: BroadcastFormProps) {
   const [isSending, setIsSending] = useState(false);
   const [title, setTitle] = useState('');
   const [message, setMessage] = useState('');
+  const titleLength = useMemo(
+    () => new TextEncoder().encode(title).length,
```
zhelezkov, line 36: minor, but maybe we shouldn't create a new text encoder instance each time?

zhelezkov (approved):
lgtm, minor q

packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
Kiryous: Good point, will update

PR #146, opened 2022-07-13
Docs/adjust docs with sdk

fsher:


zhelezkov (approved):
this is looks very good

PR #147, opened 2022-07-13
Refactor/addresses sdk

Kiryous:


packages/dialect-react-sdk/hooks/useAddresses.ts
```diff
@@ -0,0 +1,398 @@
...
+      if (!isWalletConnected || !dappPublicKey || !address) return;
+      setVerifyingCode(true);
+      try {
+        const data = await walletsApi.addresses.verify({
+          addressId: address.id,
+          code,
+        });
+        await mutateAddresses(mergeAddress(data));
+      } catch (err) {
+        throw err as Error;
```
fsher, line 334: I see a lot of similar casts to `Error`. May I ask why is that necessary?

packages/dialect-react-ui/components/Notifications/index.tsx
```diff
@@ -35,111 +37,153 @@ interface NotificationsProps {
...
 
-  const { navigate } = useRoute();
+      if (
+        !wallet?.publicKey ||
+        isFetchingAddresses ||
+        isFetchingThread ||
+        isCreatingThread ||
+        isDeletingThread
+      )
+        return;
```
fsher, line 98: could you please add a block statement for this `return`. Otherwise it looks like it's floating :D

packages/dialect-react-ui/components/Notifications/screens/Settings/CreateNotificationsThread.tsx
```diff
@@ -58,33 +52,45 @@ const CreateNotificationsThread = ({
...
+        encrypted: false,
+        backend,
       });
-  }, [backend, create, dappAddress, onThreadCreated, onThreadCreationFailed]);
+
+      // Address would be created by syncState effect
+
+      await showThread(thread);
+    } catch (e) {
+      // TODO: do we need to do smth here?
```
fsher, line 91: I'd check if `onThreadCreationFailed` was used somewhere or somehow.

fsher (approved):
Looking good, sir! Approving in advance, left minor comments

packages/dialect-react-sdk/hooks/useAddresses.ts
```diff
@@ -0,0 +1,398 @@
...
+            addressIds: addresses?.map((addr) => addr.id),
+          })
+      : null,
+    {
+      refreshInterval,
+      refreshWhenOffline: true,
+    }
+  );
+
+  const isFetchingDappAddresses =
```
zhelezkov, line 122: can you explain why is this like this?

In my mind this doesn't reflect what really happens
I understand the case you want to handle on the ui side, but in my opinion this should reflect the real process behind api

packages/dialect-react-sdk/hooks/useAddresses.ts
```diff
@@ -0,0 +1,398 @@
...
+  | ToggleParams
+  | VerifyParams
+  | ResendParams;
+
+type AddressEnriched = Address & {
+  enabled: boolean;
+  dappAddress?: DappAddress;
+};
+interface UseAddressesValue {
+  addresses: Record<AddressType, AddressEnriched> | Record<string, never>;
```
zhelezkov, line 52: why type like this?

packages/dialect-react-sdk/hooks/useThread.ts
```diff
@@ -78,11 +77,11 @@ const useThread = ({
 
   return {
     // sdk
-    thread,
+    thread: thread || null,
     delete: deleteThread,
 
     // react-lib
-    isFetchingThread,
+    isFetchingThread: thread === undefined && !errorFetchingThread,
```
zhelezkov, line 84: see comment above, wondering why it should be like this if it doesn't reflect what's really happens in the api

packages/dialect-react-ui/components/Notifications/index.tsx
zhelezkov: think we should ban this style in eslint haha

zhelezkov (changes requested):
this is good, great work. left some open questions to discuss

packages/dialect-react-sdk/hooks/useThread.ts
Kiryous: It actually reflects the loading state what happens in API, better then isValidating. See SWR docs, 

https://swr.vercel.app/docs/getting-started#make-it-reusable.

The isFetching would be true every time the request is made, it will result into flickering loading state every pollingInterval. 

What we need instead just the initial loading.

packages/dialect-react-sdk/hooks/useAddresses.ts
zhelezkov: My main concern is about `Record<string, never>`

packages/dialect-react-sdk/hooks/useAddresses.ts
```diff
@@ -0,0 +1,398 @@
...
+  refreshInterval?: number;
+}
+
+function useAddresses({
+  refreshInterval,
+}: UseAddressesParams = EMPTY_OBJ): UseAddressesValue {
+  const { wallet: walletsApi } = useDialectSdk();
+  const { dappAddress: dappPublicKey } = useDialectDapp();
+  const { connected: isWalletConnected } = useDialectWallet();
+  const [isCreatingAddress, setCreatingAddress] = useState(false);
```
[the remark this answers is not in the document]
Kiryous: hey! thanks for the proposal, I don't want to merge all this in one 'cause it could produce unexpected behaviour. E.g. while one address is deleting, another one could be created but the addressState would be set to null in success handler

packages/dialect-react-sdk/hooks/useAddresses.ts
```diff
@@ -0,0 +1,398 @@
...
+              dappAddress,
+              enabled: Boolean(dappAddress?.enabled),
+            };
+          })
+        : [],
+    [addresses, dappAddresses]
+  );
+
+  const mergeAddress = useCallback(
+    (data) =>
```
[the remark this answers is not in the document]
Kiryous: yep

packages/dialect-react-sdk/hooks/useAddresses.ts
Kiryous: @zhelezkov I added it to support empty object as value — https://stackoverflow.com/a/68208381/10197410, but prob it's not necesary

packages/dialect-react-sdk/hooks/useAddresses.ts
Kiryous: nice catch, removed it

PR #154, opened 2022-07-20
Feature/hardware wallet signin

zhelezkov:


packages/dialect-react-sdk/context/DialectContext/Sdk/index.ts
```diff
@@ -22,7 +24,33 @@ function useDialectSdk(
...
+        tokenStore = dialectCloud.tokenStore;
+      }
+
+      const token = tokenStore.get(adapter.publicKey);
+      if (!token) return;
+      const tokenValid = Auth.tokens.isValid(token);
+      if (!tokenValid) return;
+      initiateConnection();
+    },
+    [config, adapter, initiateConnection]
```
fsher, line 51: Effect is missing `dialectCloud` in `deps` array.

packages/dialect-react-ui/entities/errors/ui/NotAuthorizedError.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <Centered>
+      <H3
+        className={clsx(textStyles.header, 'dt-flex dt-items-center dt-mb-4')}
+      >
+        Verify Wallet
+      </H3>
+      <span
+        className={clsx(
+          textStyles.body,
+          'dt-flex dt-items-center dt-mb-4 dt-w-[80%] dt-text-center'
```
fsher, line 22: Why 80% width?

packages/dialect-react-sdk/context/DialectContext/Sdk/index.ts
```diff
@@ -22,7 +24,33 @@ function useDialectSdk(
...
+  useEffect(
+    function preValidateSdkToken() {
+      if (!dialectCloud) return;
+      if (!dialectCloud.tokenStore) return;
+      if (!adapter.publicKey) return;
+      // extract token store
+      let tokenStore: TokenStore;
+      if (typeof dialectCloud.tokenStore === 'string') {
+        if (dialectCloud.tokenStore !== 'local-storage') return;
+        tokenStore = TokenStore.createLocalStorage();
```
fsher, line 40: I'm not fully understanding what' going on here. Could you elaborate either here or by adding comments, please?

fsher (approved):
Well done! Left minor comments, and good to go imo!

packages/dialect-react-sdk/context/DialectContext/Sdk/index.ts
zhelezkov: yea, but it's not needed, since it has `config`, which contains dialectCloud. does it make sense?

fsher: yes, but ESLint would throw a warning here. I understand your point, but would replacing `config` with `dialectCloud` break this effect? If so, why?

packages/dialect-react-ui/entities/errors/ui/NotAuthorizedError.tsx
fsher: AFAIK, usually containers are using 2 things: max-width and horizontal paddings/margins to adjust the container placement. 80% seems like a vague number and we won't be able to accommodate to every modal size (especially, since it's configurable). Either way, your call. Just was interested why 80%. Thanks!

zhelezkov: what's your suggestion?

packages/dialect-react-sdk/context/DialectContext/Sdk/index.ts
zhelezkov: technically shouldn't break anything, replaced to `dialectCloud`

packages/dialect-react-sdk/context/DialectContext/Sdk/index.ts
zhelezkov: good point, added description, logic is a bit tricky

PR #157, opened 2022-07-24
hotfix/toggle-enabled

cbosborn:


zhelezkov (approved):
good, probably we'll revisit this piece of code during settings screen revamp

PR #160, opened 2022-07-28
feat: new settings UI

AlekseiPetrovDev:


packages/dialect-react-ui/components/Notifications/screens/NewSettings/Wallet.tsx
```diff
@@ -0,0 +1,154 @@
...
+import clsx from 'clsx';
+import { Button, Loader } from '../../../common';
+import { P } from '../../../common/preflighted';
+import { useTheme } from '../../../common/providers/DialectThemeProvider';
+import IconButton from '../../../IconButton';
+
+export const display = (publicKey: PublicKey | string): string => {
+  const s = publicKey.toString();
+  return `${s.slice(0, 4)}...${s.slice(s.length - 4)}`;
+};
```
zhelezkov, line 22: guess we have this somewhere in the utils

packages/dialect-react-ui/components/Notifications/screens/NewSettings/Wallet.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const { thread, delete: deleteDialect } = useThread({
+    findParams: { otherMembers: dappAddress ? [dappAddress] : [] },
+  });
+
+  const { delete: deleteAddress } = useAddresses();
+
+  const deleteThread = async () => {
+    await deleteDialect();
+    await deleteAddress({ addressType });
```
zhelezkov, line 63: hm, I'm not sure that we should delete address, we probably want delete dappAddress, otherwise it might cancel all other subscriptions. @tsmbl please correct me

packages/dialect-react-ui/components/Notifications/screens/NewSettings/Wallet.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+      // await showThread(thread);
+    } catch (e) {
+      // TODO: do we need to do smth here?
+    }
+  };
+
+  const { isDeletingThread } = useThread({
+    findParams: { otherMembers: dappAddress ? [dappAddress] : [] },
+  });
```
zhelezkov, line 89: looks like a hook duplicate, we have same hook above

zhelezkov (changes requested):
this is super good, minor q

packages/dialect-react-ui/components/Notifications/screens/NewSettings/Wallet.tsx
AlekseiPetrovDev: Also was a bit confused. this part was in old settings, but in general agree

packages/dialect-react-ui/components/Notifications/screens/NewSettings/Email.tsx
```diff
@@ -0,0 +1,203 @@
...
+      await createAddress({ addressType, value: email });
+      setError(null);
+    } catch (e) {
+      setError(e as Error);
+    }
+  };
+
+  const deleteEmail = async () => {
+    try {
+      await deleteAddress({ addressType });
```
zhelezkov, line 86: and one more time about deletion, guess we shouldn't delete address, but only dapp address? @tsmbl do you have insights on this?

zhelezkov (changes requested):
this is good, my only concern is about addresses deletion, @tsmbl can you help us here?

zhelezkov:
or maybe I don't know the logic behind this, and it should work like this, e.g. deleting address should disable all notifications on all other data source, but IMO this won't be clear for user

packages/dialect-react-ui/components/Notifications/screens/NewSettings/Wallet.tsx
tsmbl: If you delete address, all subscriptions are deleted across all dapp-sites
This action is pretty destructive but still a valid operation, so I expect at least to have a pop-up with warning or smth like this to communicate consequences to user

@zhelezkov @AlekseiPetrovDev wdyt?

zhelezkov: thanks for clarifying, highlighting this problem

tsmbl:
@AlekseiPetrovDev @zhelezkov where's edit address button btw?

packages/dialect-react-ui/components/Notifications/screens/NewSettings/Email.tsx
tsmbl: technically it's ok to delete it, I think, but consequences MUST be explicitly communicated to user

AlekseiPetrovDev:
> @AlekseiPetrovDev @zhelezkov where's edit address button btw?

You need to start typing something in input when address already confirmed

cbosborn (approved):
@AlekseiPetrovDev awesome work, this looks fantastic. as discussed, a couple odds & ends still need to be addressed, but nothing strictly blocking for release. going to merge now

PR #161, opened 2022-07-29
Feature/unread notifications

zhelezkov:


tsmbl (approved):
Awesome job!
btw released dialect sdk version 0.5.1 and bumped version here
I think this is good to go

PR #165, opened 2022-08-04
feature: add headless mode for <Broadcast /> component to be able to …

Kiryous:
…customise the compose Broadcast message form in future

### DIFF
--- a/packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
+++ b/packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
@@ -21,6 +21,7 @@ const ADDRESSES_REFRESH_INTERVAL = 10000;
 interface BroadcastFormProps {
   dapp: Dapp;
   headless?: boolean;
+  notificationTypeId?: string;
 }
 
 const getUsersCount = (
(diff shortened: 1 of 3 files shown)


packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
```diff
@@ -174,6 +175,11 @@ function BroadcastForm({ dapp }: BroadcastFormProps) {
   }
 
   const sendBroadcastMessage = async () => {
+    if (noUsers) {
+      setError(new Error('No users in the audience for this broadcast'));
```
zhelezkov, line 188: does it makes sense to store error but not just error msg?

zhelezkov (approved):

packages/dialect-react-ui/components/Broadcast/BroadcastForm.tsx
Kiryous: Hm.I do store error because I assume there could be some useful info in case of the real error in underlying sdk method occurred. However, I can't come up with some use case right now;

I could change to store only message. Wdyt?

PR #170, opened 2022-08-09
Fix/toggles and settings refactoring

zhelezkov:
Basically it contains fixes and refactoring which led to this:
- hook `useAddresses` split to two different hooks, `useNotificationChannel` and `useNotificationChannelDappSubscription`. First one allows to control notification channel(or address) globally. Second one is about to control dapp subscription itself. All new hooks are bounded to one specific `addressType`, e.g. wallet or telegram or email, etc..
- fixed toggler -> made them controllable and stateless
- notification mode route pick and web3 logic simplification
- removed old settings
- type fixes


packages/dialect-react-sdk/hooks/useNotificationChannel.ts
```diff
@@ -0,0 +1,197 @@
...
+          (await mutateAddresses())?.find((it) => it.type === addressType) ||
+          null
+        );
+      } finally {
+        setUpdatingAddress(false);
+      }
+    },
+    [
+      address,
+      addressType,
```
AlekseiPetrovDev, line 123: I think addressType not needed here as dependencies

packages/dialect-react-sdk/hooks/useNotificationChannel.ts
```diff
@@ -0,0 +1,197 @@
...
+      isUpdatingAddress,
+      mutateAddresses,
+      walletsApi.addresses,
```
AlekseiPetrovDev, line 126: also create/delete using walletApi object. There we get specific addresses. What's the reason of this?

AlekseiPetrovDev:
looks good to me

packages/dialect-react-sdk/hooks/useNotificationChannel.ts
zhelezkov: why, updateAddress function uses it :)

tsmbl (approved):
lgtm

packages/dialect-react-sdk/hooks/useNotificationChannel.ts
zhelezkov: eslint puts deps like this. I prefer to have whole walletApi, replaced!

PR #171, opened 2022-08-09
fix: force width and height for togglers

zhelezkov:
found that sometimes uses some css resets or tailwind form, which basically forces checkbox width to smth and it breaks our styling. The idea is to add important for togglers w/h
Screenshot 2022-08-09 at 16 47 15


AlekseiPetrovDev (commented):
lgtm

AlekseiPetrovDev (approved):

PR #175, opened 2022-08-09
fix: define `UseDappAudienceValue`, make the return type consistent

Kiryous:


packages/dialect-react-ui/hooks/useDappAudience.ts
```diff
@@ -16,19 +17,41 @@ interface UseDappAudienceParams {
...
+  telegram?: number;
+  phone?: number;
+  email?: number;
+};
+
+const DEFAULT_COUNTS: ChannelCountsType = {
+  wallet: undefined,
+  email: undefined,
+  phone: undefined,
+  telegram: undefined,
```
zhelezkov, line 31: maybe null?

zhelezkov (approved):
overall looks good, minor comment

packages/dialect-react-ui/hooks/useDappAudience.ts
Kiryous: yep, nice point

PR #177, opened 2022-08-10
upd notifications header

zhelezkov:
updates for realms use case with master back button
update stylings


AlekseiPetrovDev (approved):
lgtm

PR #181, opened 2022-08-18
removed not used styles, refactor a bit

AlekseiPetrovDev:


zhelezkov (approved):
looks like a good step to the bright future. Is this like a final clean up or some follow up coming? I thought on maybe some unification between Chat/Notifications styles. I know that chat is not in a very good shape rn, so might be out of scope

AlekseiPetrovDev:
Second part also is coming soon, but let's merge it

AlekseiPetrovDev:
Second part also is coming soon, but let's merge it

PR #183, opened 2022-08-23
Feature/unread across chats

fsher:
Reference: https://github.com/dialectlabs/sdk/pull/60

At the moment, we only mark the thread as read, but don't show it on UI.

Previously had 2 hooks, but after a discussion with @zhelezkov, it was decided to merge into 1 with a unified return type


packages/dialect-react-sdk/hooks/useGlobalUnreadMessages.ts
```diff
@@ -0,0 +1,31 @@
...
+interface UseGlobalUnreadMessagesParams {
+  refreshInterval?: number;
+}
+
+interface UseGlobalUnreadMessageValue {
+  unreadCount: number;
+  refresh: () => void;
+}
+
+const useGlobalUnreadMessages = ({
```
zhelezkov, line 15: maybe we could keep only one hook for this?
e.g. keep the logic in useUnreadMessages hook depending if public key is passed or not?

I'm worrying about it, cause we have a ton of public hooks already 😅

zhelezkov (changes requested):

packages/dialect-react-sdk/hooks/useGlobalUnreadMessages.ts
fsher: I hear you. It did seem like an unnecessary hook, given one already exists.

I initially followed the logic I used in data-service, that this seems like a separate thing, since it's "global", across all dialects. Let me try to merge them together, but I feel like we're bloating one hook.

fsher: Done, sir!

packages/dialect-react-sdk/hooks/useUnreadMessages.ts
```diff
@@ -1,37 +1,77 @@
...
 }
 
-const useUnreadMessages = ({
-  otherMembers,
-  refreshInterval,
-}: UseUnreadMessagesParams): UseUnreadMessagesValue => {
+const isParamsExtended = (
+  params: BaseUseUnreadMessagesParams | PerThreadUseUnreadMessagesParams
+): params is PerThreadUseUnreadMessagesParams =>
+  !!(params as PerThreadUseUnreadMessagesParams).otherMembers;
```
zhelezkov, line 30: very minor, but I'd use `Boolean()` instead of `!!`

zhelezkov (approved):
overall looks good

packages/dialect-react-sdk/hooks/useUnreadMessages.ts
fsher: K, done, sir!

PR #184, opened 2022-08-25
Feature/package versions on UI

fsher:
Adding versions to chat components and the notification center.

Versions rendered the following way without labels: `<react-ui version> / <react-sdk version>`

*Note*: Version in chat is "hidden" under scroll. So to see it you need to scroll down in every case

Notifications - Settings Screen
notification-version

Inbox - Threads List
chat-version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 # misc
 .DS_Store
 *.pem
+/packages/**/version.ts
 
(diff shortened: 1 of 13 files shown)


zhelezkov (approved):
LOOKS SOLID!!! GTG!!

PR #186, opened 2022-08-26
Feature/fix polling for chat

fsher:
Fixes non-existent polling in chat.

Values are arbitrary, 2s for general polling, 5s for connectivity (but might be even unnecessary, since we poll more often.
As an option, add some logic into calculating the proper polling interval for connectivity, something like:
```
if (DEFAULT_CONNECTIVITY_POLLING_INTERVAL > pollingInterval) {
 // use polling interval or don't poll at all
} else {
 // use the default, so that we know about connectivity issues early on
}
```

Thoughts?

### DIFF
--- a/packages/dialect-react-ui/components/Chat/screens/Main/ThreadsList/index.tsx
+++ b/packages/dialect-react-ui/components/Chat/screens/Main/ThreadsList/index.tsx
@@ -13,6 +13,7 @@ import { useTheme } from '../../../../common/providers/DialectThemeProvider';
 import { useRoute } from '../../../../common/providers/Router';
 import MessagePreview from './MessagePreview';
 import { UI_VERSION } from '../../../../../version';
+import { useChatInternal } from '../../../provider';
 
 interface ThreadsListProps {
   onThreadClick?: (dialectAccount: Thread) => void;
@@ -22,8 +23,8 @@ const ThreadsList = ({ onThreadClick }: ThreadsListProps) => {
   const {
     params: { threadId },
   } = useRoute<{ threadId?: ThreadId }>();
-  const { threads } = useThreads();
-  // const threads = [];
+  const { pollingInterval } = useChatInternal();
+  const { threads } = useThreads({ refreshInterval: pollingInterval });
   const {
     info: { apiAvailability },
   } = useDialectSdk();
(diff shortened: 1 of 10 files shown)


packages/dialect-react-ui/components/Chat/screens/Main/ThreadsList/index.tsx
```diff
@@ -22,8 +23,8 @@ const ThreadsList = ({ onThreadClick }: ThreadsListProps) => {
   const {
     params: { threadId },
   } = useRoute<{ threadId?: ThreadId }>();
-  const { threads } = useThreads();
-  // const threads = [];
+  const { pollingInterval } = useChatInternal();
+  const { threads } = useThreads({ refreshInterval: pollingInterval });
```
Kiryous, line 27: IMO, the polling interval could live in `dialectConfig`, so you could set the default value once and it's will be used without props-drilling and inside `react-sdk`, not `react-ui`. Wdyt?

packages/dialect-react-ui/entities/wrappers/ConnectionWrapper.tsx
```diff
@@ -29,8 +35,8 @@ export default function ConnectionWrapper({
   } = useDialectConnectionInfo();
 
   // FIXME: trigger some fetch from some of backends to get connection state
-  const { errorFetchingThreads } = useThreads({
-    refreshInterval: 5000,
+  useThreads({
```
Kiryous, line 38: Did you test it, does it produce error screen in case of network error?

Kiryous (approved):
Looks good! However, I have a few conceptual questions. Please, let's discuss them. Approving in advance in case it's blocking some feature.

packages/dialect-react-ui/entities/wrappers/ConnectionWrapper.tsx
fsher: No, I haven't explicitly, but I've only removed the unused variable. If it's necessary to make a request, this would do one, wouldn't it?

packages/dialect-react-ui/components/Chat/screens/Main/ThreadsList/index.tsx
fsher: As discussed offline, polling is more related to UI, rather than the SDK. We don't want it to be global for all SWR hooks. 

Currently the only prop-drilling that's happening is from the very high level components to the the `ChatProvider`. After that we access the context in necessary components and use it granularly.

So for now, leaving as is.

PR #187, opened 2022-08-26
[wip]Feature/per channel broadcast

Kiryous:
The PR introduces utility function


packages/dialect-react-ui/utils/toAddressType.ts
```diff
@@ -0,0 +1,21 @@
+import { AddressType, IllegalStateError } from '@dialectlabs/react-sdk';
+
+export type PersistedAddressType = 'email' | 'wallet' | 'telegram' | 'sms';
+
+const persistedAddressTypeToAddressType: Record<
+  PersistedAddressType,
+  AddressType
+> = {
+  ['email']: AddressType.Email,
```
fsher, line 9: I think you can omit the square brackets here. So:
```
= {
  "email": AddressType.Email,
  ...
}
```

packages/dialect-react-ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dialectlabs/react-ui",
-  "version": "1.0.0-beta.49",
+  "version": "1.0.0-beta.50",
```
fsher, line 3: I would recommend bumping the version only after a release (so once merged to `master` and actually released). And the changes in the Changelog to be under the `UNRELEASED` section before actual release in order to avoid confusion.

fsher (approved):
looks good, although I don't understand yet, how is this being used

I've left some minor comments!

PR #189, opened 2022-08-31
feat: added international number support

AlekseiPetrovDev:


zhelezkov (commented):
this is all good, but how do we feel about it? :)
https://bundlephobia.com/package/react-phone-number-input@3.2.11

@fsher @Kiryous your thoughts are welcome here
I don't have a strong preference here, but leaning more into that it's too much to just bring a formatting for phone numbers, wdyt?

Screenshot 2022-08-31 at 22 39 51

AlekseiPetrovDev:
I also thought about this, I can try to find formatter only without UI libs. or it could be simplified our custom solution. So open question

fsher:
I agree with @zhelezkov on this one. Feels way to large of a package for something not that impactful.

And it feels like we can utilize regexp and `replace` here, as shown in a example here: https://stackoverflow.com/questions/8358084/regular-expression-to-reformat-a-us-phone-number-in-javascript

This solution looks tiny and relies on a proper regexp, we just need to write one and that would save us a lot of bundle size.

Ideally we can cover with a couple of small unit tests, but that's optional.

I would personally insist on keeping the bundle size as small as possible, and it seems that this is one of the examples where we should not use a library to do this for us.

Happy to discuss further.

fsher:
If we don't want to waste too much time on this, there is another package that seems much smaller, and simple as a stick (but might require React plugin to be installed as well): https://bundlephobia.com/package/imask@6.4.2

But again, if we can do it on our own, I would try to do it.

zhelezkov:
@fsher unfortunately regexp and imask is not enough to match phone number. Phone numbers with international code are much trickier than it seems, this solution 100% might result in incorrect formatting

just thinking: maybe we can omit formatting at all? 😂

fsher:
what if we do 2 inputs? first - pick country code, second the actual phone number?

Kiryous:
@fsher 
>what if we do 2 inputs? first - pick country code, second the actual phone number?
This is exactly what https://catamphetamine.gitlab.io/react-phone-number-input/ is doing. I don't think we want to re-write the whole logic by ourselfs.

Kiryous:
@zhelezkov 
>just thinking: maybe we can omit formatting at all? 😂
I'm with you on this. It's better to unblock this pretty requested feature right away.

cc @AlekseiPetrovDev

fsher:
if it adds 44kb mgziped and I have a requirement that bundle size should be small, I'd rewrite it.

My main point, there should be a way that doesn't add more to the bundle, since it's clearly an issue for people integrating Dialect, that doesn't add that much to a product and that we cannot live without it. We can always as @zhelezkov suggested do a simple validation, like `+/d{1,3}/d{6,7}` or something like that.

AlekseiPetrovDev:
So I think it that case let's omit this at all. 

Because any library will add some size to bundle. From scratch this feature looks like not couple of hours work, because there is could be really tricky cases. 

Validation for numbers and special symbol like (+) should be enough then

Wdyt @fsher @zhelezkov @Kiryous ?

fsher:
I'm happy, agreed!

zhelezkov:
agree!

zhelezkov (approved):

Kiryous (approved):
LGTM!

PR #195, opened 2022-09-09
fix: `useDappAudience()` if there're no notifications type set up for…

Kiryous:
… dapp, return all unique users count

Context in 40 secs: https://www.loom.com/share/1f438207f096494abb97937473f63b19?from_recorder=1&focus_title=1

### DIFF
--- a/packages/dialect-react-ui/utils/addressesUtils.ts
+++ b/packages/dialect-react-ui/utils/addressesUtils.ts
@@ -33,13 +33,11 @@ export const getUsersCount = (
 
   // If there're no notifications type set up for this dapp, return all unique users count
   if (!subscriptions.length && notificationTypeId === null) {
-    return [
-      ...new Set(
-        enabledAndVerifiedAddresses.map((addr) =>
-          addr.address.wallet.publicKey.toBase58()
-        )
-      ),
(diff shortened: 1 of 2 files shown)


packages/dialect-react-ui/utils/addressesUtils.ts
```diff
@@ -31,9 +31,15 @@ export const getUsersCount = (
     .filter((address) => address.address.verified)
     .filter((address) => addressTypePredicate(address.address.type));
 
-  // If there're no notifications type set up for this dapp, return all addresses count
+  // If there're no notifications type set up for this dapp, return all unique users count
   if (!subscriptions.length && notificationTypeId === null) {
-    return enabledAndVerifiedAddresses.length;
+    return [
+      ...new Set(
```
fsher, line 37: `Array.from(new Set(...)).length` is also good.

fsher (approved):
LGTM!

packages/dialect-react-ui/utils/addressesUtils.ts
fsher: Or even better: `new Set().size`. So no need to convert to array to get the length.

PR #202, opened 2022-09-20
fix/get deduplication id from send command

prsvic:
tro avoid duplication when resending failed message


fsher (approved):
Great catch! LGTM

PR #211, opened 2022-11-03
[CORE-577]feat: add blockchainType into useDapp hook

dshuvalov:
# Feature
- add blockchainType into useDapp hook


zhelezkov (approved):
lgtm gtg lfg

PR #229, opened 2023-06-24
feat: add ability to inject custom components into channels (both rem…

fsher:
…ote and local)

Adds a `renderAdditional` property to `NotificationType` type and a new `remoteNotificationExtension` prop that would find the remote notification by matching the `humanReadableId` (TBD, needs to be clarified how unique is this) and pass down `renderAdditional` property.

Screenshot 2023-06-24 at 19 03 32

### DIFF
--- a/examples/notifications-solana/pages/index.tsx
+++ b/examples/notifications-solana/pages/index.tsx
@@ -116,7 +116,7 @@ export default function Home(): JSX.Element {
 
   const dialectConfig = useMemo((): ConfigProps => {
     return {
-      environment: 'development',
+      environment: 'production',
       dialectCloud: {
         tokenStore: 'local-storage',
       },
--- a/examples/notifications-solana/pages/index.tsx
+++ b/examples/notifications-solana/pages/index.tsx
@@ -66,8 +66,12 @@ function AuthedHome() {
             notifications={[
               {
                 name: 'Welcome Message',
-                detail:
-                  'Welcome message that is sent on first subscription',
+                detail: 'Welcome message that is sent on first subscription',
+                renderAdditional: () => (
+                  <div className="text-xs text-neutral-400">
+                    Additional information about the welcome message
+                  </div>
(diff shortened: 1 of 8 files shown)


zhelezkov (approved):

examples/notifications-solana/pages/index.tsx
```diff
@@ -66,8 +66,12 @@ function AuthedHome() {
             notifications={[
               {
                 name: 'Welcome Message',
-                detail:
-                  'Welcome message that is sent on first subscription',
+                detail: 'Welcome message that is sent on first subscription',
+                renderAdditional: () => (
```
Kiryous, line 70: let's add optional argument 'state' to the example. something like

```
renderAdditional: (state: boolean) => (
    <div className="text-xs text-neutral-400">
        Additional information about the welcome message (now {state ? 'turned on' : 'turned off'})
    </div>
    )
```

packages/dialect-react-ui/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [UNRELEASED]
 
+- feat: adds `renderAdditional` render property for `NotificationType` type which allows to render custom content in notification settings
+- feat: adds `remoteNotificationExtensions` prop that adds custom content to remote notifications
```
Kiryous, line 6: let's add example of usage of `remoteNotificationExtensions`

Kiryous (approved):
LGTM, minor suggestions around example

examples/notifications-solana/pages/index.tsx
fsher: Since these are local notifications, there is no toggle, which means there is no state.

I agree that we need an example for remote notifications, but we don't have any for now.

zhelezkov (approved):

Kiryous (approved):
LGTM

PR #235, opened 2024-04-22
fix/thread summary polling if not found

prsvic:
Problem: to determine if user has unread notifications, we check /summary, which returns 404 if thread does not exist. It creates a lot of spam errors in console, sentry etc.

Solution: stop polling once 404 is met ('data' returned by sdk is null)
When user enables notifications for wallet, refresh 'data' value in swr to reset polling interval.


packages/react-sdk/src/hooks/index.ts
```diff
@@ -14,3 +14,4 @@ export { default as useThread } from './useThread';
 export { default as useThreadMessages } from './useThreadMessages';
 export { default as useThreads } from './useThreads';
 export { default as useUnreadMessages } from './useUnreadMessages';
```
zhelezkov, line 16: do we need this hook or we can delete it?

zhelezkov (approved):
looks good, let's decide if we can remove the old hook

